PR #3, opened 2018-06-19
Add a Flask template

ajm348:
I'm not really sure if this is how it's supposed to look/be, so feel free to tear this apart


flask/files/.platform/services.yaml
```diff
@@ -0,0 +1,6 @@
+mysql:
+  type: "mysql:10.1"
+  disk: 2048
+
+redis:
+  type: "redis:3.0"
```
Crell: Why not 3.2?

flask/files/.platform/services.yaml
```diff
@@ -0,0 +1,6 @@
+mysql:
+  type: "mysql:10.1"
```
Crell: Why not 10.2?

flask/files/README.md
```diff
@@ -0,0 +1,24 @@
+# Platform.sh Python 2.7 minimal example
```
Crell: This file looks like a copy-paste from the minimal Python repo.  It should be updated to say that it's Flask.

dodo.py
```diff
@@ -285,7 +325,7 @@ def common_branch(root):
 def common_push(root):
     return {
         'actions': [
-            'cd %s/template && git checkout update && git push -u origin update' % root,
+            'cd %s/template && git checkout -B update && git push -u origin update' % root,
```
Crell: What's the reason for this change?

README.md
```diff
@@ -0,0 +1,21 @@
+# Utility scripts to manage the Platform.sh Template projects.
```
Crell: Yay, better README!  

No period at the end of titles, though.  Or colons on the others.  Also, leave one blank line between the header and the content below it.

Crell:
Do you not have access to put the branch in the repo directly?  It's harder to collaborate in a fork.

dodo.py
ajm348: It was because the Flask repo did not have a branch called `update`. Likely unnecessary now that the branch does exist, right? I can remove it.

ajm348:
I don't have access to push to the repo directly.

Crell:
You now have access to push to this repo. :smile:

flask/files/README.md
Crell: PR every time.  The goal is to hook those PRs up to Psh test environments so we can review the changes before merging them.

dodo.py
Crell: The branch should be created in the `common_branch()` step.  The push step assumes there's already an update branch to be pushed; if there isn't then adding the `-B` here will make an empty PR, as the new branch will be exactly the same as master.

At least that's how I think it should all work. :smile:

pjcdawkins:
Curious - why python 2.7 and not 3.6?

ajm348:
Ok, I updated versions for redis & mysql, and Crell finally pushed "how it's supposed to work" through my thick skull via slack. Why Python 2.7 vs 3.6... I don't have a good answer for that @pjcdawkins . Personal bias I guess?

Crell:
Thanks, @ajm348!  Merged.

PR #5, opened 2018-06-22
Drupal8 Fix

Crell:
Fixes issues generating all the necessary files, and updates to the latest changes from the example project.


Crell:
Um, why is this conflicting?

PR #9, opened 2018-10-01
Magento2 CE

Crell:
A working Magento 2 CE builder!  Based very closely on work by @ralt.  I'm mainly just converting it into the template-builder format here.

Open question: I am still strongly considering renaming `deploy` to `deploy.py` and running it directly with `python deploy.py` rather than relying on the file being executable.  It feels more self-documenting and robust that way, and 98% of the time users won't know or care about the difference.


magento2ce/files/README.md
```diff
@@ -0,0 +1,26 @@
...
+3. Run the provided Git commands to add a Platform.sh remote and push the code to the Platform.sh repository.
+
+That's it!  Your Magento site will deploy and you'll be able to view it in a browser.  Note that as part of the setup process the admin account is already created for you.  The initial username `admin` has a password of `admin123`.  You will be required to change it the first time you login (at `/admin` in the browser).
+
+## Using as a reference
+
+If you have an existing Magento 2 CE site you need to prepare for running on Platform.sh, the following files are the most important:
+
+* The [`.platform.app.yaml`](/.platform.app.yaml) file and the [`.platform`](/.platform) directory, which tell the system how to assemble your containers.
+* The [`deploy`](/deploy) script, which is a Python script that checks to see if Magento needs to be installed and installs it if necessary.
```
ralt: The `deploy` script (no need to specify "Python" imo...) does more than the install: it also runs db upgrades on a normal deployment.

magento2ce/files/README.md
```diff
@@ -0,0 +1,26 @@
...
+* The [`app/etc/local.xml`](/.app/etc/local.xml) file.
+* The [`pub/static-versioned.php`](/.pub/static-version.php) file, which provides an alternative front-controller for static files.
```
ralt: URL is wrong there

ralt: Also, the "alternative" word sounds wrong there? As in, it's not going to work without it.

magento2ce/files/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+There is also a [patch file](https://github.com/platformsh/template-builder/blob/master/magento2ce/platformsh.patch) needed to make the installer run correctly.  It may not be necessary if your site is already setup and you're just migrating it.
```
ralt: The patch is removing stupid stuff that Magento is doing anyway, so it doesn't hurt to apply it. I'd remove the last sentence.

dodo.py
```diff
@@ -353,7 +390,7 @@ def common_update(root, branch):
             'cd %s/template && git checkout master' % root,
             'cd %s/template && git fetch --all --depth=2' % root,
             'cd %s/template && git merge --allow-unrelated-histories -X theirs --squash project/%s' % (root, branch),
-            'cd %s/template && composer update --no-interaction' % root
+            'cd %s/template && composer update --ignore-platform-reqs --no-interaction' % root
```
ralt: That argument sounds like you're asking for our customers to have trouble on production.

Crell: That's only run while building the template repo locally (on one of our computers).  Since the resulting container for Magento is still built on 7.1 and runs without issue it doesn't seem to cause a problem.  It's just allowing it to run `composer install` on any PHP version for the initial creation.

magento2ce/files/README.md
Crell: Yes, but it's alternative to default Magento behavior, no?

PR #11, opened 2018-10-10
Add a gitkeep file to force the Drupal config/sync directory to exist.

Crell:
cf https://github.com/platformsh/template-drupal8/issues/14


pjcdawkins:
It also needs to be configured in settings.php

Crell:
It is: https://github.com/platformsh/template-builder/blob/master/drupal8/files/web/sites/default/settings.php#L33

PR #18, opened 2018-11-29
Add Django 2

mcyprian:
Adding Django 2 template and doit tasks to manage it easily.


django2/files/.platform.app.yaml
```diff
@@ -0,0 +1,62 @@
...
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+    database: "postgresqldb:postgresql"
+
+# The configuration of app when it is exposed to the web.
+web:
+    # Whether your app should speak to the webserver via TCP or Unix socket
+    # https://docs.platform.sh/configuration/app-containers.html#upstream
+    upstream:
+        socket_family: unix
```
Crell: I'm surprised by this.  All of the other examples we have use a port, not a unix socket.  Is there a reason for it?  Is this in the current example repo this way?

django2/files/routes.yaml
```diff
@@ -0,0 +1,13 @@
+# The routes of the project.
```
Crell: This file needs to be inside the `.platform` directory under `files`.

django2/files/services.yaml
```diff
@@ -0,0 +1,10 @@
+# The services of the project.
```
Crell: This file needs to be inside the `.platform` directory under `files`.

django2/files/routes.yaml
mcyprian: Oh, I have copied it to the wrong path. fixed.

django2/files/.platform.app.yaml
mcyprian: I took this as a working code snippet from Django 1.X example without special inspection. It seems using `-b 0.0.0.0:$PORT` as a gunicorn flag works equally well. [Official Django docs page](https://docs.djangoproject.com/en/2.1/howto/deployment/wsgi/gunicorn/) also doesn't mention anything about unix sockets. Maybe we can ask original author @JGrubb if it can be simply replaced by `gunicorn -w 4 -b 0.0.0.0:$PORT myapp.wsgi:application`.

JGrubb: Funny enough, I just started a Django project last night so I'll try using a port instead.

@Crell to your point though - we likely didn't have any/enough documentation on using sockets vs ports at the time this was put together.  Most of the Python and Ruby examples used to tend toward sockets instead of ports for whatever reason.

Crell: If a port works equally well, I'd marginally favor switching to that for consistency's sake.  I always had trouble setting up socket listeners when making templates, but I suspect it was just a question of who built the template originally.  (The Ruby examples are mostly Ori, for instance.)

Crell:
This seems fine to me after the rebase, although cleaning that up took a bit.  Not merging yet while we wait to hear from @JGrubb on the socket/port question.

JGrubb:
I have no real dog in this hunt and your "keep it consistent" is a virtue worth pursuing.  You have my blessing...

PR #21, opened 2018-12-07
Add template files

alphon-sh:


django2/files/.platform.template.yaml
```diff
@@ -0,0 +1,40 @@
+## Project behavior schema
+
+# Allows products and project creation to define
+# various settings and workflow modifications
```
Crell: This header tells me nothing.  This file is going to end up in my repository when I create a project, and I'm going to wonder WTF it is.  The header needs to be very clear on what it's for, and the fact that I can delete it.

(Which... I kind of think means this is a bad place to put it but I know Ori feels strongly about it.)

Crell:
I've updated the descriptions on most projects, as discussed.

Crell:
Template file headers have been updated per discussion in Slack.

PR #24, opened 2018-12-12
Laravel - updating readme to include changes we made

omurphy27:
1) Pointing out the phpredis patch that we made

2) Also making it more clear that redis is being used by default as Laravel's cache and session driver (vanilla Laravel uses a different default for these).


Crell:
This will now need a rebuild of the Laravel template to make it public.  Remember the hassle that is...  (Still don't have a good solution there.)

PR #31, opened 2018-12-20
Use doit sub-tasks

Crell:
Making a PR for review purposes...


mcyprian:
This MR is ready for the review. The dodo.py script won't need to be changed at all when adding a new template. I also created a hierarchy of classes to be able to overwrite individual tasks. Please be pedantic when reviewing this, these changes required a lot of moving and copy pasting so some regretions might appear in the code.

project/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+            name = self.name
+        return ['git clone git@github.com:platformsh/template-{0}.git {1}'.format(
+            name, self.builddir)
+        ]
+
+    @property
+    def platformify(self):
+        actions = ['rsync -aP {0} {1}'.format(
+            os.path.join(TEMPLATEDIR, self.name, 'files'),  self.builddir
+        )]
```
Crell: This appears to be rsyncing from the wrong source.  See: https://github.com/platformsh/template-drupal8/pull/23/files

It looks like the `files` directory is included in the content that gets synced.  That probably means we need a `/` on the `files` string in the join, because rsync is super picky about its slashes.

Crell: I confirmed in testing that just adding a / seems to do the job.  I think. :smile:

mcyprian: Fixed, resulting actions list for platformify command now looks the same as in the master branch.

mcyprian:
@Crell feel free to add docs and comments. I will check if it matches my intentions and understanding of refactored structure and layout of the project.

mcyprian:
All doc updates make sense to me. It was a pleasure collaborating with you.

Crell:
Likewise!  Thank you for all your help, and may your future adventures be enjoyable and profitable.

PR #34, opened 2019-02-04
Language updates

Crell:
Old code: Auto-update composer on all remote projects, even if they weren't PHP.

New code: Auto-update each language's package manager if the file is there, for all projects.

### DIFF
--- a/project/__init__.py
+++ b/project/__init__.py
@@ -19,6 +19,7 @@ class BaseProject(object):
     # project build directory to update the corresponding lock file.
     updateCommands = {
         'composer.json': 'composer update --prefer-dist --ignore-platform-reqs --no-interaction',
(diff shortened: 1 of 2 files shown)


project/__init__.py
```diff
@@ -14,6 +14,14 @@ class BaseProject(object):
     to override.
     '''
 
+    # A dictionary of conditional commands to run for package updaters.
+    # The key is a file name. If that file exists, then its value will be run in the
+    # project build directory to update the corresponding lock file.
+    updateCommands = {
+        'composer.json': 'composer update --prefer-dist --ignore-platform-reqs --no-interaction',
+        'Pipfile': 'pipenv update',
+    }
```
Crell: Update this with the appropriate Node and Ruby commands.

PR #35, opened 2019-02-13
Remove host filtering in Drupal, as Platform.sh already does it. 

Crell:
This should also be applied to the Drupal 7 repos before we commit.


pjcdawkins:
The change is good 👍 ... I don't understand the "before we commit" so I'm not merging, but FWIW I don't think Drupal 7 has an equivalent (just `$base_url`, which we don't set)

chadwcarlson:
@Crell got verification that host header is also replaced on Enterprise, so we can move forward with these.

Crell:
w00t!

PR #38, opened 2019-03-04
Adds the GovCMS 8 template

xtfer:
This adds a GovCMS 8 template based on the Drupal 8 template. The main difference is a custom composer.json to pull govCMS.

Built template located at https://github.com/platformsh/template-govcms8.

Functional site built from template at https://master-7rqtwti-ij33tnfffx4i4.au.platformsh.site.


Crell:
My only significant pushback is "why is this GovCMS 8"?  Is it 8 as in "Drupal 8", or v8 of GovCMS?  We try not to include versions in the repo as we only want to ship the latest version of each to begin with.  (Drupal 7 and Drupal 8 are an exception only because they're so different that they should really be considered separate applications to begin with.)

What happens when GovCMS 8.4 or 9 are released?

xtfer:
@Crell It's GovCMS in Drupal 8, as opposed to govCMS in Drupal 7. It is known as GovCMS 8.

Crell:
Ah ha, OK.  That makes sense then.

Looking it over again, though, it looks like you're including a revised composer.json file.  It should instead use the upstream composer.json (so that we can get updates to it), and in the GovCMS8 class run a `composer require` command to add the additional govcms dependency.  (I *think* that's the only difference?)  The `Symfony4` class is a good example of that you can copy-paste-modify.

xtfer:
@Crell Updated with changes as requested.

Crell:
I needed to add the Govcms8 class to the imports in doit.py, but otherwise, looks good.  THanks, @xtfer!

PR #39, opened 2019-03-04
Move Wordpress root to /web, as per original build.

xtfer:
This moves the Wordpress root to /web (not /web/wp) as per the original example. This will fix the issue with wp-content directory paths being incorrect, and is a better solution than routing those paths into web/wp, as it allows local development to occur without rebuilds on every code change.


templates/wordpress/files/composer.json
```diff
@@ -0,0 +1,22 @@
+{
```
Crell: This file comes from upstream so we shouldn't be overwriting it.  Modifying via script only.

templates/wordpress/files/web/index.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
Crell: Does this come from upstream, or is it vanilla ours?

dodo.py
```diff
@@ -12,7 +12,6 @@
 from project.laravel import Laravel
 from project.magento import Magento2ce
 from project.symfony import Symfony3, Symfony4
-from project.wordpress import Wordpress
```
Crell: It's not going to work without the class available, methinks...

templates/wordpress/files/composer.json
xtfer: I don't understand why we do this. Using one composer file to pull another dependency with a composer file is design intent of Composer. In this case, we set the relevant install paths for the upstream in our file. This is safer than overwriting the source, IMHO.

templates/wordpress/files/web/index.php
xtfer: It's ours. Moving the wp web root doesn't work without a custom index.php

PR #42, opened 2019-03-25
Automatically dereive the latest tag for a remote project to merge from

Crell:
I've tested this against Laravel, but not the other projects yet.  But it does remove an annoying bit of hard coding in a nice way.

Also, we have to keep upstream_branch as the Drupal Composer upstream project... has no tags.  It's just an 8.x branch that wraps whatever the latest tagged version of Drupal is.  So... *sigh*

Drupal 7 is also an oddball since that needs updates to the makefile, not just to the definition class.  That hasn't been addressed yet here.

### DIFF
--- a/project/laravel.py
+++ b/project/laravel.py
@@ -2,7 +2,7 @@ from .remote import RemoteProject
 
 
 class Laravel(RemoteProject):
-    upstream_tag = 'v5.8.3'
+    major_version = 'v5'
     remote = 'https://github.com/laravel/laravel.git'
 
     @property
--- a/project/magento.py
+++ b/project/magento.py
@@ -2,5 +2,5 @@ from .remote import RemoteProject
 
 
 class Magento2ce(RemoteProject):
-    upstream_branch = '2.2'
+    major_version = '2.2'
     remote = 'https://github.com/magento/magento2.git'
--- a/project/remote.py
+++ b/project/remote.py
@@ -38,17 +38,13 @@ class RemoteProject(BaseProject):
                 subprocess.check_output('cd {0} && git merge --allow-unrelated-histories -X theirs --squash {1}'.format(
                 self.builddir, latest_tag), shell=True)
             actions.append(merge_from_upstream_tag)
-        elif hasattr(self, 'upstream_tag'):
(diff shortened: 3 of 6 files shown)


chadwcarlson:
Initial Testing (3/25/2019):
- Laravel: Fine
- Magento2CE: fails in `platformify`
`Hunk #1 FAILED at 316.
Hunk #2 succeeded at 454 (offset 2 lines).
Hunk #3 succeeded at 811 (offset 17 lines).
1 out of 3 hunks FAILED -- saving rejects to file setup/src/Magento/Setup/Model/Installer.php.rej
TaskFailed - taskid:platformify:magento2ce
Command failed: 'cd /Users/chadcarlson/Documents/active/template-builder/template-builder/templates/magento2ce/build/ && patch -p1 < /Users/chadcarlson/Documents/active/template-builder/template-builder/templates/magento2ce/platformsh.patch' returned 1
`
- Symfony3&4: fails in `update`. `Exception: No upstream tag found to merge from`
`TaskError - taskid:update:symfony3
PythonAction Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/doit/action.py", line 424, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/Users/chadcarlson/Documents/active/template-builder/template-builder/project/remote.py", line 37, in merge_from_upstream_tag
    latest_tag = self.latest_tag()
  File "/Users/chadcarlson/Documents/active/template-builder/template-builder/project/remote.py", line 65, in latest_tag
    raise Exception('No upstream tag found to merge from')
Exception: No upstream tag found to merge from`
- Wordpress: Fine, but no environment build test set up

Crell:
Magento was that the patch file didn't apply anymore.  I fixed that.

Symfony was that they use a v prefix on tags and I didn't catch that.  Fixed the major version marker on those.

Wordpress: Skip for now, because it needs a major overhaul to begin with.  (The PR for it adds 1700 files, meaning something else is very off.)

Crell:
Magento 2 CE update here as well: https://github.com/platformsh/template-magento2ce/pull/6

PR #43, opened 2019-03-27
Use Config Reader for Drupal 8

Crell:
This is built off of the autotag branch to reduce the chances of conflicts, which may or may not be the best decision but meh...

Ready for review.  Don't merge until autotag is in, of course.


pjcdawkins:
👍 for the settings.platformsh.php changes

except I think we wanted to stop looping through routes, instead having something like:

```php
// The Host header can trusted on Platform.sh servers.
$settings['trusted_host_patterns'] = ['.*'];
```

..because this has rightly moved from being an "example" (here's what you could do if you wanted) to being a "template" (here's exactly what we think you should do).

PR #52, opened 2019-04-18
Add Pyramid template.

chadwcarlson:
add python pyramid


templates/pyramid/files/.platform.app.yaml
```diff
@@ -0,0 +1,35 @@
...
+      pipenv install --system --deploy
+# The size of the persistent disk of the application (in MB).
+disk: 1024
+# The relationships of the application with services or other applications.
+#
+# The left-hand side is the name of the relationship as it will be exposed
+# to the application in the PLATFORM_RELATIONSHIPS variable. The right-hand
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+    mysql: "mysql:mysql"
```
Crell: Almost everywhere else we call the main DB relationship `database`.  Let's stick to that.

templates/pyramid/files/.platform/services.yaml
```diff
@@ -0,0 +1,6 @@
+mysql:
```
Crell: I've been trying to avoid calling the mysql service `mysql`, as then you just end up with mysql repeated over and over. :smile: I think I've generally called it `mysqldb`.

templates/pyramid/files/app.py
```diff
@@ -0,0 +1,16 @@
+import os
```
Crell: The `services.yaml` file includes both MySQL and Redis, but neither are being used, or even have demo code commented out.  Is there anything we can provide in that regard?  (I don't know Python/Pyramid well enough to say what that would even look like.)

templates/pyramid/files/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+pyramid = "*"
```
Crell: Should this also include config-reader, so the user doesn't have to think about that later when they start using a database?

templates/pyramid/files/app.py
```diff
@@ -0,0 +1,16 @@
...
+
+def hello_world(request):
+    return Response('Hello World!')
+
+if __name__ == '__main__':
+    with Configurator() as config:
+        config.add_route('hello', '/')
+        config.add_view(hello_world, route_name='hello')
+        app = config.make_wsgi_app()
+    server = make_server('127.0.0.1', int(os.environ["PORT"]), app)
```
Crell: This should be using config-reader, no?

templates/pyramid/files/Pipfile
```diff
@@ -0,0 +1,15 @@
...
+platformshconfig = "*"
```
Crell: I don't know if this is typical in Python-land, but I'm a little concerned with this self-updating itself to v3 of the library if we ever release a v3 with a BC break. Is that something we should be concerned about?

PR #56, opened 2019-04-23
Add Config-reader

chadwcarlson:
Small Readme change to Python 2. Add config-reader to Python3 and Flask


templates/flask/files/.platform.template.yaml
```diff
@@ -36,6 +36,7 @@ info:
   notes:
     - heading: "Apps & Services"
       content: "Python 2.7<br/>MySQL 10.2<br/>Redis 3.2"
+    - heading: "Requirements"
```
Crell: Hm, we just removed this line from the template files.  Merge issue or old data, maybe?

PR #59, opened 2019-04-26
Python template fixes and basic template naming updates

chadwcarlson:
Flask - update requirements to remove security warning from repo
Basic Python 2 - De-flask, fix dependency issue with `requirements.txt` for now, template update name
Basic Python 3 - De-flask, add config-reader, update template name
Django 1 - fix dependency issue with `requirements.txt` for now
Django 2 - add some config-reader in `settings.py`, `ROUTES` and `RELATIONSHIPS` don't work for now with config-reader, since `settings.py` is called in build.
Python 3 UWSGI - add config reader, added service tests to use it
Basic PHP - update template name 
Basic Go - update template name


templates/golang/files/.gitignore
```diff
@@ -1,2 +1,3 @@
 /.GOPATH
 /bin
+.platform/local/*
```
Crell: Shouldn't the local directory have its own .gitignore in it that works?  This shouldn't be necessary.  (Let's figure out why it was necessary and make it not necessary.)

templates/python3/files/.platform.template.yaml
```diff
@@ -39,7 +39,7 @@ info:
 # This key describes the initialization call made to the master environment at
 # project creation time. This is part of the full v2 UI operation mode, which
 # places project schema/options selection early in the creation process, rather
-# than later as it exitss now. To allow this schema to be backwards-compatible,
+# than later as it exits now. To allow this schema to be backwards-compatible,
```
Crell: "exists" :-)

For a follow-up: "as it exists now" is no longer true since the new UI is deployed, so we probably want to fix this block across the board at some point.

PR #63, opened 2019-05-02
Add Wagtail Django CMS template.

chadwcarlson:


templates/wagtail/files/myapp/wsgi.py
```diff
@@ -0,0 +1,16 @@
...
+
+For more information on this file, see
+https://docs.djangoproject.com/en/2.1/howto/deployment/wsgi/
+"""
+
+import os
+
+from django.core.wsgi import get_wsgi_application
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'myapp.settings')
```
Crell: Is "myapp" a Wagtail thing, or something we're doing?

chadwcarlson: It's taken from the Django 2 template naming, which does seem to be a Django naming convention at least for examples.

templates/wagtail/files/.platform.app.yaml
```diff
@@ -0,0 +1,63 @@
...
+  # assembling the codebase.
+  build: |
+    pipenv install --system --deploy
+    pip install wagtail
+    
+    mkdir logs
+    python manage.py collectstatic
+    rm -rf logs
+  deploy: |
+    python manage.py migrate
```
Crell: This is safe to do on every push, without overwriting existing data?

chadwcarlson: Ya this checks for differences, and is consistent with our other templates: https://github.com/platformsh/template-django2/blob/762d73a762d1a27391eac35e84b4565e2f633acb/.platform.app.yaml#L61

templates/wagtail/files/myapp/wsgi.py
Crell: Hm.  Not a fan, but we have nothing better I suppose.

PR #70, opened 2019-05-13
Template pimcore5

RibanKeke:
Hello Dev_rel. Here is my fresh pull request. 
Waiting for your remarks and feedbacks.


project/pimcore.py
```diff
@@ -0,0 +1,14 @@
+from .remote import RemoteProject
+
+
+class Pimcore5(RemoteProject):
+    major_version = 'v5.8.2'
```
Crell: This shouldn't pin to an exact version.  Then it never auto-updates.  It should pin to `v5` if we want it to automatically pull in the latest 5.x version every time we run update, or to `v5.8` if we want it to only auto-update with the 5.8 series.

I don't know which is appropriate for Pimcore, but I'd assume `v5` is what we want?

project/pimcore.py
```diff
@@ -0,0 +1,14 @@
...
+    remote = 'https://github.com/pimcore/pimcore.git'
+
+    @property
+    def platformify(self):
+        return ['cd {0} && rm -rf .ci .travis .gitlab-ci.yml .travis.yml .github CONTRIBUTING.md'.format(self.builddir),
```
Crell: We don't strip these files in any other repo.  No need to do so here.  If we were to start doing that we should do it in some more centralized and automated way.

project/pimcore.py
```diff
@@ -0,0 +1,14 @@
...
+                'cd {0} && COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/skeleton tmp_build'.format(self.builddir),
+                'cp -r {0}tmp_build/* {0} && rm -r {0}tmp_build'.format(self.builddir) 
+                ] + super(Pimcore5, self).platformify + [
+                'cd {0} && composer update --no-scripts && composer require platformsh/config-reader doctrine/orm --no-scripts'.format(self.builddir),]
```
Crell: Interesting.  This would be the only time we put the parent commands in the middle of the series.  Why?

project/pimcore.py
```diff
@@ -0,0 +1,14 @@
+from .remote import RemoteProject
+
+
+class Pimcore5(RemoteProject):
+    major_version = 'v5.8.2'
+    remote = 'https://github.com/pimcore/pimcore.git'
```
Crell: I think this is the issue.  If we're building off of pimcore/skeleton, then pimcore/skeleton is what should be listed as the remote.  That's all composer create-project does anyway.  That also then removes the need for the funky tmp_build directory in platformify.

project/pimcore.py
Crell: Is doctrine/orm not included in the skeleton out of the box?  Why is that a separate require?

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+# The left-hand side is the name of the relationship as it will be exposed
+# to the application in the PLATFORM_RELATIONSHIPS variable. The right-hand
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+  database: "pimcore_mysqldb:pimcore_admin"   
+  redis: 'rediscache:redis'
+
+# Add additional PHP extensions.
+build:
+  flavor: none
```
Crell: Why?  The command in the build hook below looks like the default, or close enough.  Just use the composer build flavor, or if there's a reason not to include that in a comment in the build hook.

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+# The relationships of the application with services or other applications.
+#
+# The left-hand side is the name of the relationship as it will be exposed
+# to the application in the PLATFORM_RELATIONSHIPS variable. The right-hand
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+  database: "pimcore_mysqldb:pimcore_admin"   
+  redis: 'rediscache:redis'
+
+# Add additional PHP extensions.
```
Crell: Wrong comment.  Or rather, it belongs on the runtime block below.

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - opcache
+    - imagick
+    - fileinfo
+
+variables:
+  env:
+    PIMCORE_INSTALL_ADMIN_USERNAME: "admin"
+    PIMCORE_INSTALL_ADMIN_PASSWORD: "admin"
+  php:
+      memory_limit: 512M
```
Crell: Dear god, what an obscenely wasteful application.  Is that necessary?

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - simplexml
+    - exif
+    - opcache
+    - imagick
+    - fileinfo
+
+variables:
+  env:
+    PIMCORE_INSTALL_ADMIN_USERNAME: "admin"
+    PIMCORE_INSTALL_ADMIN_PASSWORD: "admin"
```
Crell: These worry me considerably.  Does this have any impact after the initial install?  If not, let's include a comment that users should remove those after installation.  If it does, then we can't have them here at all because it's a security hole.

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+
+# The 'mounts' describe writable, persistent filesystem mounts in the application. The keys are
+# directory paths, relative to the application root. The values are strings such as
+# 'shared:files/PATH', where PATH is a relative path under the mount's source directory.
+mounts:
+    "/var": "shared:files/var"
+    "/web/var": "shared:files/web-var"
+    "/pimcore": "shared:files/pimcore"
+    "/web/pimcore": "shared:files/web-pimcore"
```
Crell: For non-Drupal, we should probably start moving to the new-style mount syntax.  The only reason we use the old style in the Drupal examples is for compatibility with PE's auto-mount.  I don't think that's going to matter much longer, and definitely not for non-Drupal.

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+      root: web
+      passthru: "/app.php"
+      allow: true
+      rules:
+        ^/videos/(?<resource>.*)$:
+          passthru: "/var/assets/videos/$resource"
+        ^/img/(?<resource>.*)$:
+          passthru: "/var/assets/img/$resource"
+        ^/static/(?<resource>.*)$:
+          passthru: "/web/static/$resource"
```
Crell: Is `passthru` really the right property here?  Normally that's for mapping to a PHP file in case of file-not-found, but the links here are not PHP files.  What's the intent here?

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+disk: 2048
+
+# The hooks executed at various points in the lifecycle of the application.
+hooks:
+  build: |
+    set -e
+    composer install --no-ansi --no-progress --prefer-dist --no-scripts
+  deploy: |
+    if [ ! -f web/var/.platform.installed ]; then
+        bin/console doctrine:query:sql "ALTER DATABASE  CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;"
```
Crell: The application really ought to be doing this itself.  It's dangerous to rely on the server setting.  Have we filed an upstream bug to set the connection character set rather than relying on the DB setting?

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
...
+        if [ $? -eq 0 ]; then
+          echo "Database updated to utf8mb4_unicode_ci"
+        else
+           exit 1
+        fi
+        vendor/bin/pimcore-install --no-interaction --ignore-existing-config --no-debug
+        if [ $? -eq 0 ]; then
```
Crell: I like this.  We should probably add such error handling to other templates that do this sort of touch-file install.  (I'm not sure if the ones we maintain do so, but eZ at least does.)

templates/pimcore5/files/.platform.template.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+  # Additional notes displayed in the template's detail view.
+  # Each note object is displayed as a small section heading with content below. Supports limited HTML.
+  notes:
+    - heading: "Apps & Services"
+      content: "PHP 7.2<br/>MySQL 10.2<br/>Redis 5.0"
```
Crell: The .platform.app.yaml file says 7.3.  Let's say the same here.

templates/pimcore5/files/.platform/routes.yaml
```diff
@@ -0,0 +1,14 @@
+# The routes of the project.
+#
+# Each route describes how an incoming URL is going
+# to be processed by Platform.sh.
+
+"https://{default}/":
+    type: upstream
+    upstream: "pimcore:http"
+    cache:
+      enabled: false
```
Crell: Don't disable the cache.  If anything it should be configured properly to whitelist their session cookie(s).

templates/pimcore5/files/.platform.app.yaml
```diff
@@ -0,0 +1,100 @@
+# This file describes an application. You can have multiple applications
+# in the same project.
+#
+# See https://docs.platform.sh/user_guide/reference/platform-app-yaml.html
+
+# The name of this app. Must be unique within a project.
+
+
+# The runtime the application uses.
+name: pimcore
```
Crell: All template app containers should be named "app".

templates/pimcore5/files/.platform/services.yaml
```diff
@@ -0,0 +1,18 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+pimcore_mysqldb:
```
Crell: Just name it `mysqldb`.  Making it custom just makes code samples harder. :smile:

templates/pimcore5/files/.platform/services.yaml
```diff
@@ -0,0 +1,18 @@
...
+  type: mysql:10.2
+  disk: 2048
+  configuration:
+    schemas:
+      - project_database
+    endpoints:
+      pimcore_admin:
+        default_schema: project_database
+        privileges:
+          project_database: admin
```
Crell: None of this custom configuration should be necessary, unless Pimcore only works with that database name.  (In which case it would be very dumb.)  It can probably just be entirely omitted.  (If there's a specific reason it has to be explicit, that needs to be documented in comments.)

templates/pimcore5/files/app/config/parameters_platformsh.php
```diff
@@ -0,0 +1,31 @@
...
+if ($platformsh->hasRelationship('redis')) {
+    $redis_credentials = $platformsh->credentials('redis');
+    $container->setParameter('redis_host', $redis_credentials['host']);
+    $container->setParameter('redis_port', $redis_credentials['port']);
+}
+
+// Set a default unique secret, based on a project-specific entropy value.
+if ($platformsh->projectEntropy) {
+    $container->setParameter('kernel.secret', $platformsh->projectEntropy);
+}
```
Crell: Newline at end of file.

project/pimcore.py
RibanKeke: Doctrine is not included and it's needed for the deploy hook when updating the database encoding to the Picmore supported format.

project/pimcore.py
RibanKeke: Good catch, using the skeleton simplifies a lot the process.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: I had to add the --no-scripts option so the custom scripts are not triggered on build time, they are executed on deploy.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: https://pimcore.com/docs/5.x/Development_Documentation/Installation_and_Upgrade/System_Requirements.html they suggest more or equal to 128MB, 512MB be an overkill I'll bring it down to 128MB as suggested or 256MB, I have no experience with php about the memory usage.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: It has no impact after initial install, I'll add a comment explaining that it should be removed after the setup.

project/pimcore.py
Crell: So the Pimcore skeleton doesn't include Doctrine, but it requires doctrine?  I don't get it.

templates/pimcore5/files/.platform.app.yaml
Crell: Most well-behaved modern systems shouldn't need 512 MB, ever.  A well-behaved system will probably only ever need 20-30 MB, but will periodically spike over 100.  A 128 MB limit is probably fine.  We can easily bump it to 256 in the future if needs be.

templates/pimcore5/files/.platform.app.yaml
Crell: I'd check with the security team as well and get their signoff.  I don't want to ship this without their OK.

project/pimcore.py
RibanKeke: So to clarify, the application by itself doesn't require the doctrine/orm but requires utf8mb4 database charactet_set , so I added the doctrine/orm so I can modify the database character_set in the deploy hook, I could install the CLI and do it through the CLI too. Is there a way to do this through the service.yaml file, if not this could be handy feature to add.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: Indeed, setting the memory ton 128MB then.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: So, thinking about this. I can take example on the wagtail template and explain in the template README.md that the user has to log in SSH and execute the install script where he has to set the admin user_name and the admin_password. Anyway, if it's not done, accessing the site will show you an explicit message saying that pimcore is not installed.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: I wasn't aware of the new style mount syntax. I'll apply the changes.

templates/pimcore5/files/.platform.app.yaml
RibanKeke: You mean this should be set by the application. The character set I mean. I checked the pimcore documentation and this setting is part of the requirements for the application. Indeed I can file a bug for this but I'm not sure that they see it as a bug.

templates/pimcore5/files/.platform/services.yaml
RibanKeke: Indeed they are not needed.

templates/pimcore5/files/.platform.app.yaml
Crell: They may not, but we should file it as one anyway because it is one. :smile: The current code is needlessly brittle.

project/pimcore.py
Crell: If it's not using Doctrine post-install, using Doctrine just to issue a call against the database is way overkill.  Especially ORM, which is the full object relational mapper.  If this is just for your own script then just use PDO directly, no dependency at all.

Although it's still an upstream bug if this is required at all and Pimcore needs to fix that.

templates/pimcore5/files/.platform.app.yaml
Crell: :+1:

templates/pimcore5/files/.platform.app.yaml
dephekt: As long as the behavior is as @RibanKeke describes and is documented as described, I think that is an acceptable approach to this.

templates/pimcore5/readme.patch
```diff
@@ -0,0 +1,63 @@
+diff --git a/README.md b/README.md
```
Crell: The readme file should be just a standalone file that totally replaces the Pimcore one.  Making it a patch just complicates things.

Crell:
I successfully ran a build, but when installed it shows an error that Pimcore itself didn't install properly:

https://pr-1-djjnuwy-zkc7lgjy4ov6i.eu-3.platformsh.site/

Crell:
It looks like this is the erroring code, in `app.php`:

```php
// redirect to installer if pimcore is not installed
if (!is_file(\Pimcore\Config::locateConfigFile('system.php'))) {
    Debug::enable(E_ALL, true);
    throw new RuntimeException('Pimcore is not installed! Please install via command line.');
}
```

(No comment about the comment...)

`app/config/pimcore/system.php` is indeed missing.  I am not clear on when it's supposed to have been created.

templates/pimcore5/readme.patch
RibanKeke: Done, I pushed changes for this.

RibanKeke:
> I successfully ran a build, but when installed it shows an error that Pimcore itself didn't install properly:
> 
> https://pr-1-djjnuwy-zkc7lgjy4ov6i.eu-3.platformsh.site/

The update of the database encoding and the installation procedure are discribed in the README file. I This is no longer done through a script. The was a configuration file missing, I have added that. 
With the latest build all should worK

Crell:
I have fixed the conflict with the master branch; simple list conflict.  But deploying that still failed.  I deactivated and reactivated the pr-1 environment to make sure it was starting from a clean slate and it's still claiming that Pimcore is not installed.

RibanKeke:
@Crell 
Hi, I have went through the template again and noticed that compared to my initial commit my changes following review broke som stuff:
- The install of the doctrine/orm is required as this is used to load the platform database configuration
- The generated file paremeters.yaml has to be added manually as this i checked by the install process, the secret is generated by our platfromsh paramerters file, so I removed it.
- After the push on platform, you have to update the database encoding with the CLI and ssh on the container and start the install process with the admin credentials, all that is explained in the README file.

PR #74, opened 2019-05-22
Add simple Hugo template

chadwcarlson:
See also: https://github.com/platformsh/template-hugo/pull/1


templates/hugo/files/.gitmodules
```diff
@@ -0,0 +1,4 @@
+
+[submodule "themes/minimal"]
```
Crell: No submodules.  We do not give customers submodules.  They're a time bomb waiting to go off, usually sooner rather than later.

templates/hugo/files/.platform.app.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+# The configuration of the application when it is exposed to the web.
+web:
+    locations:
+        '/':
+            # The public directory of the application relative to its root.
+            root: 'public'
+            index: ['index.html']
+            scripts: false
+            allow: true
```
Crell: Git wants a newline.

templates/hugo/files/.platform.app.yaml
```diff
@@ -0,0 +1,26 @@
...
+name: 'hugo'
+
+# The type key specifies the language and version for your application.
+type: 'golang:1.12'
+
+# The hooks that will be triggered when the package is deployed.
+hooks:
+    # Build hooks can modify the application files on disk but not access any services like databases.
+    build: |
+      ./build.sh
```
Crell: We can go ahead and use the !include syntax here.  It should work for plain text still, even if archive is broken.

templates/hugo/files/.platform.app.yaml
```diff
@@ -0,0 +1,26 @@
+# .platform.app.yaml
+
+# The name of this application, which must be unique within a project.
+name: 'hugo'
```
Crell: App name is always `app` in examples/templates.

templates/hugo/files/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+HUGOVERSION=0.55.6
+
+wget https://github.com/gohugoio/hugo/releases/download/v$HUGOVERSION/hugo_${HUGOVERSION}_Linux-64bit.tar.gz
+tar xvzf hugo_${HUGOVERSION}_Linux-64bit.tar.gz
+
+rm hugo_${HUGOVERSION}_Linux-64bit.tar.gz
+
+mv ~/themes/minimal/exampleSite/content/post/* ~/content/post/
+mv ~/themes/minimal/exampleSite/content/project/* ~/content/project/
```
Crell: This speaks to the ongoing question of whether this should be a starting point or an example, I believe.  We don't have a definitive answer for that yet, other than "we'll want some of both".  Which this should be is unclear.

Crell: I'm confused though.  Why are we pulling in content via submodule, AND copying it around, AND we have it included by default below, for two different themes?  I don't understand how that works.

chadwcarlson: All this is doing is copying in `exampleSite` pages from the `minimal` theme into `app/content`. 

As far as I can tell, since this example content is not in the parent `content` directory originally, it will not be a part of the site since it is in the theme directory. This was a simple submodule that provided some example content, but not all of the ones I looked at had any. There's only one theme though.

If that didn't happen (or if we don't want to do that) there are `posts/hello-world.md` and `projects/hello-world.md` that would be the only content in those two dirs.

chadwcarlson: The example content provided by the theme just explains how to edit when using that theme (e.g. code block syntax), which to me seemed like a starting point of "here's a working hugo template with this theme, here's where your files are, and here's how you should edit and modify them."

Crell: If the goal is to be a starting point, then that's probably better.  Don't give people even more pages they have to remove, let them just start adding content.

PR #75, opened 2019-05-28
Add instructions for contributing new templates.

chadwcarlson:


README.md
```diff
@@ -63,3 +63,72 @@ And poof, you are ready to make a PR with the updates.
...
+
+
+## Contributing
+
+If you would like to contribute to the list of Platform.sh's maintained templates, there is a simple process for ensuring that future maintenance is set up from the start through `template-builder`.
+
+Let's use the previous example: you have created a new application that uses the framework Spiffy that you think would be a useful template. 
+
+1. Create a repository under the [Platform.sh Github organization](https://github.com/platformsh/) for the template. Officially maintained templates begin with the `template-` prefix, so name it `template-spiffy`, and initialize it with an empty `README.md`.
+2. Add the "Examples and Templates team" as `Admin` collaborators in the `platformsh/template-spiffy` repository settings.
```
Crell: I think I usually use Write, not Admin.  But it probably should be Admin, actually.

README.md
```diff
@@ -63,3 +63,72 @@ And poof, you are ready to make a PR with the updates.
...
+3. Set up GitHub integration for the repository. When you submit your template for review it will helpful if there is already a live site that shows what the application would look like to customers and that it deploys properly. You can find information about how to set up a [GitHub integration](https://docs.platform.sh/administration/integrations/github.html#github) in the documentation.
```
Crell: This is missing a step to create a Psh project, and what region it should be on (EU-3).  Also, this would be a good time to discuss whose key to use for these integrations.  We should probably not always use mine. :smile:

README.md
```diff
@@ -63,3 +63,72 @@ And poof, you are ready to make a PR with the updates.
...
+4. Clone the [`template-builder` repository](https://github.com/platformsh/template-builder) locally. Create and checkout a new branch called `add-spiffy`.
+5. Each template project is in its own directory within `templates`, which corresponds to a GitHub repository with the `template-` prefix. Create the directories `templates/spiffy` and `templates/spiffy/files/` on the `add-spiffy` branch. 
+6. Copy the files you want to be in `template-spiffy` into `templates/spiffy/files/`. If your application requires any patches to deploy on Platform.sh, copy them into `templates/spiffy/`.
```
Crell: Since I give it 50/50 odds that the reader hasn't read the first half of the page, this point should emphasize that it's *just* the files that need to get *added*, and that the build process in Python should do as much of the work as possible, via downloading dependences, linking to an upstream, etc.  How much of that goes here vs deep-linking to code comments or the text above is debatable.

README.md
```diff
@@ -63,3 +63,72 @@ And poof, you are ready to make a PR with the updates.
...
+7. Include or update the `README.md` so that it is similar to other templates, and include any information specific to running the application on Platform.sh you think the customer should know.
+8. Each template comes with a file called `.platform.templates.yaml`, which is used to define how the template repository will appear in and initialize from the management console. For example:
+
+    ```yaml
```
Crell: No need to inline the whole thing.  Just deep-link to one.

README.md
chadwcarlson: I think I started using `Admin` because I saw it on another repo, but maybe I'm wrong. Let's set in stone what we want though.

README.md
chadwcarlson: I guess I left that open for the moment because we hadn't decided/set up a separate account for templates yet, suggesting that new submissions should at the very least be set up on a project on THEIR account for the review process. Then we could delete the integration and substitute it for something else when we do have that set up. 

Although maybe now's the time. How long it would it take to set up a `templates@platform.sh` account?

README.md
```diff
@@ -63,3 +63,33 @@ And poof, you are ready to make a PR with the updates.
...
+
+Let's use the previous example: you have created a new application that uses the framework Spiffy that you think would be a useful template. 
+
+1. Create a repository under the [Platform.sh Github organization](https://github.com/platformsh/) for the template. Officially maintained templates begin with the `template-` prefix, so name it `template-spiffy`, and initialize it with an empty `README.md`.
+2. Add the "Examples and Templates team" as `Admin` collaborators in the `platformsh/template-spiffy` repository settings.
+3. Set up GitHub integration for the repository. When you submit your template for review it will helpful if there is already a live site that shows what the application would look like to customers and that it deploys properly. You can find information about how to set up a [GitHub integration](https://docs.platform.sh/administration/integrations/github.html#github) in the documentation. Create a blank project on your Platform.sh account called `template-spiffy` on the region "Europe (West 3)" (`eu-3.platform.sh`) and set up the integration.
```
Crell: This step can now be updated to assign an appropriate owner.

PR #79, opened 2019-06-10
Adds two new Java templates

otaviojava:


templates/template-spring-boot-maven-mysql/files/.platform/local/README.txt
```diff
@@ -0,0 +1,8 @@
+.platform/local
```
chadwcarlson: Need to ignore everything in .platform/local so it doesn't push to the template repos.

templates/template-spring-mvc-maven-mongodb/files/.platform/local/README.txt
```diff
@@ -0,0 +1,8 @@
+.platform/local
```
chadwcarlson: Same note as above.

templates/template-spring-mvc-maven-mongodb/files/README.md
```diff
@@ -0,0 +1,19 @@
+# Spring MVC Maven template for Platform.sh
+
+This project provides a starter kit for Spring Boot Maven with MySQL projects hosted on [Platform.sh](http://platform.sh).
```
chadwcarlson: I missed this in the last PR, but these READMEs are identical. This doesn't address MVC and it still says MySQL instead of Mongo.

chadwcarlson:
The Mongo PR needs to be updated in the actual repo as well, sorry I didn't catch that first. 

Another thing, the individual template directories (i.e. templates/template-spring-mvc-maven-mongodb) should have the same format as the other templates in there (https://github.com/platformsh/template-builder/tree/add_java_template/templates). Don't include the `template-` prefix in the directory names, especially if there is some logic in template-builder that assumes that it wont be included wrt the actual template repository.

PR #84, opened 2019-06-18
Skeleton for OpenProject

Crell:
This is still very broken.

### DIFF
--- a/templates/openproject/files/.platform.app.yaml
+++ b/templates/openproject/files/.platform.app.yaml
@@ -1,0 +1,20 @@
+name: "app"
+
+type: "ruby:2.6"
+disk: 1024
+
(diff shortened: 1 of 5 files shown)


templates/openproject/files/.platform.app.yaml
```diff
@@ -0,0 +1,20 @@
...
+        rails db:migrate
+
+mounts:
+    "/log": "shared:files/log"
+    "/tmp": "shared:files/tmp"
+    "/db": "shared:files/db"
+
+web:
+    commands:
+        start: 'rails server -p $PORT'
```
OriPekelman: You should not run in production the built-in server. Production use requires an application server such as Unicorn or Puma

PR #88, opened 2019-06-20
Add Beego template.

chadwcarlson:
https://github.com/platformsh/template-beego/pull/1


templates/beego/files/conf/app.conf
```diff
@@ -0,0 +1,4 @@
+appname = template-beego
+//httpaddr = "127.0.0.1"
+httpport = 8888
```
Crell: Can this be made dynamic?  It's always 8888 on Psh right now, but it ought to be `$PORT`.

chadwcarlson: I was trying to do that before when I was asking if PORT was available at build. My thinking:

- I tried `httpport = $PORT`, and when it didn't work I thought "Okay, .conf files can read env vars, I guess I should write this file on build."
- Tried doing that, PORT not available in build. 
- left it as hardcoded 8888

How else should I approach this?

templates/beego/files/controllers/default.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/astaxie/beego"
+)
+
+type MainController struct {
+	beego.Controller
+}
+
+func (c *MainController) Get() {
+	c.Data["Website"] = "beego.me"
+	c.Data["Email"] = "astaxie@gmail.com"
```
Crell: Who is astaxie, and why are we trying to spam him?

chadwcarlson: I can remove that, this is a part of the standard Welcome to Beego page. I can remove everything but the website.

Crell:
This seems fine to me otherwise, aside from being very bare.  I don't know if we want to figure out the "right" way to hook a DB into this or not.  Since it's a micro-framework, essentially, maybe not?

templates/beego/files/conf/app.conf
Crell: Does the author have a help channel, or project, or forum, or something where we can ask?  Doing it in a way that plays nice with the framework would be ideal.  If there isn't one, then I guess we hard code it with a comment about why it's being hard coded.

templates/beego/files/controllers/default.go
Crell: I don't know why he wants so much spam, but yeah, let's remove the email line at least to be good netizens. :smile:

templates/beego/files/conf/app.conf
```diff
@@ -0,0 +1,4 @@
+appname = template-beego
+//httpaddr = "127.0.0.1"
+httpport = 8080
```
Crell: Ours is 8888.  Or is this not being read?

templates/beego/files/controllers/services.go
```diff
@@ -0,0 +1,139 @@
+package controllers
+
+import (
+        "fmt"
+        "database/sql"
+        _ "github.com/go-sql-driver/mysql"
```
Crell: Does the _ import here not cause problems, as it did on Gin?

chadwcarlson: No none of the same problems that I could see.

templates/beego/files/main.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/astaxie/beego"
+	"github.com/platformsh/template-beego/controllers"
+	_ "github.com/platformsh/template-beego/routers"
+)
+
+func getPort() string {
+	var c controllers.MainController
+	c.Prepare()
+	return c.Config.Port()
+}
```
Crell: Clever!  I don't know if it's too clever, but definitely clever. :smile:

chadwcarlson: I don't know either, but it works!

templates/beego/files/controllers/services.go
```diff
@@ -0,0 +1,139 @@
...
+  if err != nil {
+    panic(err)
+  }
+
+  affect, err = res.RowsAffected()
+  if err != nil {
+    panic(err)
+  }
+
+  db.Close()
```
Crell: I realized when poking at the language-examples version, this line should probably be moved to `defer db.Close()` right after the Open() call's error handling.  That's the more Go-ish way of making sure things get closed.  (Not a merge blocker, but something we should probably change across all the examples.)

chadwcarlson: Like this?

```
  db, err := sql.Open("mysql", formatted)
  if err != nil {
    panic(err)
  }
defer db.Close()
```

templates/beego/files/routers/router.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/platformsh/template-beego/controllers"
+	"github.com/astaxie/beego"
+)
+
+func init() {
+    beego.Router("/", &controllers.MainController{})
+		beego.Router("/test", &controllers.MySQLTestController{})
+}
```
Crell: If I'm following correctly, this means the import statement in main.go automatically triggers this init() to run, and that's what does the wiring to configure Beego, correct?  Because otherwise I don't see how the configuration happens.  Is this how Beego documents you should do it?

templates/beego/files/controllers/services.go
```diff
@@ -0,0 +1,141 @@
...
+  if err != nil {
+    panic(err)
+  }
+
+  affect, err = res.RowsAffected()
+  if err != nil {
+    panic(err)
+  }
+
+  // db.Close()
```
Crell: This line can be removed.

templates/beego/files/conf/conf.go
```diff
@@ -0,0 +1,16 @@
+package conf
+
+import (
+  psh "github.com/platformsh/config-reader-go/v2"
+)
+
+var Config = createConfigObj()
```
Crell: Assuming people are going to use this file for other things, this should be named less generically.  PshConfig or similar.  (And then update the factory function and references to it.)

templates/beego/files/conf/conf.go
```diff
@@ -0,0 +1,16 @@
...
+
+func createConfigObj() *psh.RuntimeConfig {
+  // The Config Reader library provides Platform.sh environment information mapped to Go structs.
+  config, err := psh.NewRuntimeConfig()
+  if err != nil {
+  	panic("Not in a Platform.sh Environment.")
```
Crell: `panic(err)`.  I discovered the hard way that the err may not actually be for not being in a Psh environment, and we shouldn't suppress the actual error message.

templates/beego/files/conf/conf.go
Crell: Also, I have no idea which is more Go-ish; making this call directly here, or defining the variable and then populating it from `init()`.  Absent knowledge of that I can go either way, but calling it out in case you've seen some hint.

templates/beego/files/controllers/services.go
```diff
@@ -0,0 +1,139 @@
...
+  if err != nil {
+    panic(err)
+  }
+
+  affect, err := res.RowsAffected()
+  if err != nil {
+    panic(err)
+  }
+
+  fmt.Println(affect)
```
Crell: Stray debugging line?

PR #89, opened 2019-06-20
Revel

chadwcarlson:
https://github.com/platformsh/template-revel/pull/1


templates/revel/files/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+2. Create a new project through the Platform.sh user interface and select "Import an existing project" when prompted.
+
+3. Run the provided Git commands to add a Platform.sh remote and push the code to the Platform.sh repository.
+
+That's it!  You now have a working "hello world" level project you can build on.
+
+## Structure
+
+This project relies on Go module support in Go 1.11 and later.  You should commit your `go.mod` and `go.sum` files to Git, but not the `vendor` directory.
```
Crell: Except it's not, it looks like.  It's doing it the old-school way, which... does it even work on Psh at all? :smile: I could only barely make GOPATH work on Psh.

Crell:
My only real concern here is the not-really-modules part.  I am surprised that it works at all under those circumstances.

templates/revel/files/README.md
chadwcarlson: This one took me the longest by far to get running.

chadwcarlson: This template also has the hardcoded httpport = 8888 in it, so that should be updated as well if you can suggest a fix in https://github.com/platformsh/template-builder/pull/88

PR #92, opened 2019-06-21
Add Gin template.

chadwcarlson:
https://github.com/platformsh/template-gin/pull/1


templates/gin/files/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/platformsh/template-gin
+
+require (
+	github.com/gin-gonic/gin v1.4.0
+	github.com/go-sql-driver/mysql v1.4.1
+	github.com/platformsh/config-reader-go/v2 v2.0.1
+	google.golang.org/appengine v1.6.1 // indirect
```
Crell: indirect meaning it was added automatically?  Is... Google App Engine really a library that needs to be installed when running on Psh?  Or does this automatically get omitted at compile time if we don't use it?  I have so many questions (mostly about Google making proprietary assumptions...)

templates/gin/files/server.go
```diff
@@ -0,0 +1,147 @@
...
+
+func setupRouter() *gin.Engine {
+	// Disable Console Color
+	// gin.DisableConsoleColor()
+	r := gin.Default()
+
+	// Hello World
+	r.GET("/", func(c *gin.Context) {
+
+		config, err := psh.NewRuntimeConfig()
```
Crell: I don't know what the Go convention is for it, but we should only be instantiating this once, ideally.  Probably as a (gag) global that gets instantiated once in `init()`.

chadwcarlson: Ya I wasn't sure how to do that and had meant to come back to it when the config reader stuff came up.

templates/gin/files/go.mod
Crell: Hm, OK, it looks like that lib is much smaller than I feared, it just shares a name with Google's weak competitor to us. :smile:  Ignore this point.

chadwcarlson: Ya when you run `go mod tidy` it finds all the indirects and adds them for you. Lots of questions on my end as well.

templates/gin/files/server.go
```diff
@@ -0,0 +1,143 @@
...
+	"database/sql"
+	"fmt"
+	"github.com/gin-gonic/gin"
+	_ "github.com/go-sql-driver/mysql"
+	psh "github.com/platformsh/config-reader-go/v2"
+	sqldsn "github.com/platformsh/config-reader-go/v2/sqldsn"
+	"net/http"
+)
+
+var db = make(map[string]string)
```
Crell: Why is this a global?

And... it's a string->string hash map?  I don't get it...

templates/gin/files/server.go
```diff
@@ -0,0 +1,143 @@
...
+
+func setupRouter(config *psh.RuntimeConfig) *gin.Engine {
+	// Disable Console Color
+	// gin.DisableConsoleColor()
```
Crell: Is this still needed?  What does it do?

templates/gin/files/server.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/gin-gonic/gin"
+	_ "github.com/go-sql-driver/mysql"
+	psh "github.com/platformsh/config-reader-go/v2"
+	sqldsn "github.com/platformsh/config-reader-go/v2/sqldsn"
+	"net/http"
+)
+
+var db = make(map[string]string)
+
+func setupRouter(config *psh.RuntimeConfig) *gin.Engine {
```
Crell: This trick of passing the config to setupRouter will work for here, but it won't work for larger applications, I think.  It may or may not work for the language-examples site.  That's... probably OK for now so not a merge blocker, but we may want to refactor this in the future.

PR #100, opened 2019-07-02
fixing mounts issue in .platform.app.yaml for Laravel template

ollie-platform:
Looks like the mounts were incorrectly defined, which was causing the file cache to fail (template was still working because we were using redis for cache by default).

See related discussion from Slack here: https://platformsh.slack.com/archives/C0JHEUHQD/p1561998987313800?thread_ts=1561997839.311200&cid=C0JHEUHQD

See PR made to the Laravel Template as well: https://github.com/platformsh/template-laravel/pull/20


ollie-platform:
Opened a new PR with this fix and another: https://github.com/platformsh/template-builder/pull/103

PR #106, opened 2019-07-25
Adds Express and Koa templates.

chadwcarlson:
https://github.com/platformsh/template-express/pull/1
https://github.com/platformsh/template-koa/pull/1


templates/express/.platform.template.yaml
```diff
@@ -0,0 +1,53 @@
...
+# Templates are a small amount of information supporting a template URL. Each template is selectable at the project-creation step.
+info:
+  # Unique machine name, prefaced by a vendor or organization identifier
+  id: platformsh/express
+  # The human-readable name of the template.
+  name: Express
+  # Human-readable descriptive text for the template. Supports limited HTML.
+  description: |
+      <p>  This template builds the Express framework for Platform.sh.  It includes a minimalist application skeleton for demonstration, but you are free to alter it as needed.</p>
+      <p>  Express is a minimalist web framework written in Node.js.</p>
```
Crell: Why the leading spaces after the open-p tag?

templates/express/files/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+        await connection.query(sql);
+
+        results[1] = `Created table "${tableName}".`;
+
+        // Insert data.
+        values = `(username, departname, created)`;
+        newRow = `('platform', 'Deploy Friday', '2019-06-17')`;
+
+        sql = `INSERT INTO ${tableName} ${values} VALUES ${newRow};`
```
Crell: This should be using a prepared statement.  Literal values belong in a query precisely never. :smile:  (Unless it's truly a single literal string for the whole query.  Although even that is debatable.)

templates/express/files/index.js
```diff
@@ -0,0 +1,115 @@
...
+            port: credentials.port,
+            user: credentials.username,
+            password: credentials.password,
+            database: credentials.path,
+        });
+
+        results[0] = `Connected to MariaDB at ${credentials.host}.`
+
+        // Add the table.
+        tableName = "userinfo";
```
Crell: This is static so just inline it.

templates/express/files/index.js
```diff
@@ -0,0 +1,115 @@
...
+        return Promise.reject(error);
+      }
+}
+
+// Define the main route.
+app.get('/', async function(req, res){
+
+  var results = [];
+
+  await mariadbTest().then(function(result) { results = result;});
```
Crell: The inlined function here can use short-function syntax, which is easier to read.  And then return the value.  Comme ca:

`let results = await mariadbTest().then((result) => result);`

templates/express/files/index.js
```diff
@@ -0,0 +1,115 @@
...
+          });
+        }
+
+        results[2] = output;
+
+        // Drop the table.
+        sql = `DROP TABLE ${tableName}`;
+        await connection.query(sql);
+
+        results[3] = `Dropped table "${tableName}".`
```
Crell: This doesn't actually test anything, since tableName is statically defined above.  We can probably skip this output.

templates/koa/.platform.template.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Templates are a small amount of information supporting a template URL. Each template is selectable at the project-creation step.
+info:
+  # Unique machine name, prefaced by a vendor or organization identifier
+  id: platformsh/koa
+  # The human-readable name of the template.
+  name: Koa
+  # Human-readable descriptive text for the template. Supports limited HTML.
+  description: |
+      <p>This template builds a Koa project on Platform.sh.</p>
+      <p>Koa is a lightweight web microframework for Node.js.</p>
```
Crell: We've been skipping the "What is this?" line, I think.

templates/express/.platform.template.yaml
Crell: The second line can be skipped.

templates/express/files/README.md
```diff
@@ -0,0 +1,23 @@
+# Express for Platform.sh
+
+This template builds the Express framework for Platform.sh.  It includes a minimalist application skeleton for demonstration, but you are free to alter it as needed.
+
+Express is a minimalist web framework written in Node.js.
```
Crell: Omit.

templates/koa/files/README.md
```diff
@@ -0,0 +1,23 @@
+# Koa for Platform.sh
+
+This template builds a Koa project on Platform.sh.
+
+Koa is a lightweight web framework for Node.js.
```
Crell: Omit.

templates/koa/files/index.js
```diff
@@ -0,0 +1,110 @@
+const mysql = require('mysql2/promise');
+const config = require("platformsh-config").config();
+const Koa = require('koa');
+
+const app = module.exports = new Koa();
+
+async function mariadbTest() {
```
Crell: All the same comments apply here as in the Express template.

templates/express/files/index.js
ralt: Or just `let results = await mariadbTest();`

templates/express/files/index.js
```diff
@@ -0,0 +1,108 @@
...
+          PRIMARY KEY (uid)
+        ) DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;`;
+
+        await connection.query(sql);
+
+        results[1] = "Created table 'userinfo'.";
+
+        // Insert data.
+        values = `(username, departname, created)`;
+        newRow = `('platform', 'Deploy Friday', '2019-06-17')`;
```
Crell: Are these now unnecessary?  (A prepared statement would still be better, IMO.)

templates/express/files/.platform/services.yaml
```diff
@@ -0,0 +1,8 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+
+mydatabase:
+    type: mysql:10.2
```
otaviojava: Maybe use the `mariadb` here instead.

templates/koa/files/.platform/services.yaml
```diff
@@ -0,0 +1,8 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+
+mydatabase:
+    type: mysql:10.2
```
otaviojava: the same here

templates/express/files/.platform/services.yaml
chadwcarlson: That's how I had it before, but it resulted in the weirdness I mentioned earlier: I can use `mariadb` in type, but I would still need to use `mysql` for the endpoint.

otaviojava: ahhh yeap, I remember that

PR #111, opened 2019-08-06
Initial Lisp template

OriPekelman:
https://github.com/platformsh/template-lisp/pull/1


templates/lisp/files/example.lisp
```diff
@@ -0,0 +1,16 @@
+(defpackage #:example
+  (:use :hunchentoot :cl-who :cl)
+  (:export main))
+
+(in-package #:example)
+
+(define-easy-handler (greet :uri "/hello") (name)
+  (with-html-output-to-string (s) (htm (:body (:h1 "hello, " (str name))))))
+
+(export 'main)
```
ralt: This line is unnecessary, line 3 already takes care of it.

PR #132, opened 2019-09-10
Add ASP.NET Core template.

chadwcarlson:
https://github.com/platformsh/template-aspnet-core/pull/1


templates/aspnet-core/files/.platform.app.yaml
```diff
@@ -0,0 +1,68 @@
...
+hooks:
+  build: |
+    set -xe
+    dotnet publish --self-contained --runtime "$DOTNET_DEFAULT_RUNTIME" --output "$PLATFORM_OUTPUT_DIR" -p:UseRazorBuildServer=false -p:UseSharedCompilation=false
+
+
+# These are writable folders mounted over the read-only image of the built application.
+# Use these for static assets and uploads.
+mounts:
+  "/wwwroot/etc": "shared:files/wwwroot/etc"
```
Crell: Let's go straight to new-style mount syntax here.

templates/aspnet-core/files/.platform/routes.yaml
```diff
@@ -0,0 +1,12 @@
...
+# Each route describes how an incoming URL is going
+# to be processed by Platform.sh.
+
+"https://{default}/":
+    type: upstream
+    upstream: "app:http"
+
+"https://www.{default}/":
+    type: redirect
+    to: "https://{default}/"
```
Crell: The PR we just merged says to have the redirect the other way around. :smile:

templates/aspnet-core/files/.platform.app.yaml
chadwcarlson: Missed that

templates/aspnet-core/files/Controllers/HomeController.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Caching.Redis;
+
+namespace PlatformshAspNetCore.Controllers
+{
+    public class HomeController : Controller
+    {
+        private readonly MyDbContext _myDbContext;
```
Crell: Is underscored property names actually a C# convention?  Most languages these days view it as an anti-pattern, AFAIK.  Let's verify.  (If that's what C# recommends, then that's what we do.)

templates/aspnet-core/files/Helpers/PlatformRelationship.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Buffers.Text;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Linq;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+
+namespace PlatformshAspNetCore.Helpers
```
Crell: :scream_cat:

templates/aspnet-core/files/Helpers/PlatformRelationship.cs
```diff
@@ -0,0 +1,92 @@
+using System;
```
Crell: Shouldn't this whole thing be in a separate config reader library?

templates/aspnet-core/files/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+                var port = Environment.GetEnvironmentVariable("PORT");
+                if (port != null && Int16.TryParse(port, out var portNum))
+                    options.ListenLocalhost(portNum);
+
+                // This won't work as is now, due to:
+                // https://go.microsoft.com/fwlink/?linkid=874850
+                // Here for future reference only.
+                var socket = Environment.GetEnvironmentVariable("SOCKET");
+                if (socket != null)
+                    options.ListenUnixSocket(socket);
```
Crell: conditionals should always have their wrapping {}.  Any convention that says otherwise is wrong. :smile:

templates/aspnet-core/files/README.md
```diff
@@ -0,0 +1,26 @@
...
+* MariaDB 10.2
+* Redis 5.0
+
+## Customizations
+
+This repository was created from an ASP.NET Core MVC template and most of it still resembles that. The following files and additions make the framework work on Platform.sh. If using this project as a reference for your own existing project, replicate the changes below to your project.
+
+* The `.platform.app.yaml`, `.platform/services.yaml`, and `.platform/routes.yaml` files have been added. These provide Platform.sh-specific configuration and are present in all projects on Platform.sh. You may customize them as you see fit.
+* The `Program.cs` file has been modified to look for Platform-specific environment variables on startup, and listen on the correct TCP port.
+* The `Starup.cs` file has been modified to
```
Crell: Modified to what?

Crell:
Why are we including all of Bootstrap here?  I can't tell where much of this code came from, and what's needed for Platform and which parts are some stock starter kit I'm going to have to dismantle.

templates/aspnet-core/files/Helpers/PlatformRelationship.cs
chadwcarlson: Eventually sure we should make one. For now isn't this sufficient, provided a name other than "helper" is used?

templates/aspnet-core/files/Controllers/HomeController.cs
chadwcarlson: At least for private properties that does seem to be the case https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties#expression-body-definitions

templates/aspnet-core/files/Helpers/PlatformRelationship.cs
chadwcarlson: Evil "Helper" renamed!

templates/aspnet-core/files/Data/MyDbContext.cs
```diff
@@ -0,0 +1,15 @@
+using Microsoft.EntityFrameworkCore;
+
+namespace PlatformshAspNetCore.Data
+{
+    public class MyDbContext : DbContext
```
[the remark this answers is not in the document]
chadwcarlson: I think in this case this is extending `DbContext` from `Microsoft.EntityFrameworkCore` into the `PlatformshAspNetCore` namespace so it can be used by the application. This could be done in `Startup.cs`, but I think this is the recommended way to separate things in case you wanted to modify `MyDbContext`.

chadwcarlson:
@Crell You're right there wrt bootstrap. Removed everything from `wwwroot/lib` that was unnecessary for the basic app.

templates/aspnet-core/files/Controllers/HomeController.cs
Crell: Silly C#...

PR #134, opened 2019-09-11
Refine Drush URL selection

pjcdawkins:
- Fix upstream name match to accommodate the `:http` suffix
- Prefer the primary route
- Sort URLs rather than the list of preferred "original URLs"


Crell:
Closing in favor of #138

PR #135, opened 2019-09-13
Reorganise imports for project-builder

xtfer:
I was heartily confused by the build process, and when digging into the imports it was difficult to follow how this application was working. This PR does two things then:

1) It reorganises the imports so they make more sense and are more Pythonic:
- they are absolute (not relative)
- sorted into logical files (not declared in __init__.py, for example) and packages
- conform to PEP standards for file organisation
- does not declare helper functions in the doit file
- reduces the use of global namespace variables (ALL_PROJECTS)
- and moves any remaining variables into a tidy little config.py file

2) Adds a new doit `build` task that falls between `update` and `platformify`. This is mostly to allow build steps to occur prior to the platformify. One example is `composer require`, which should really be run prior to platformify doing its final steps. Currently it is doing nothing, however an update of platformify steps could follow. It also makes it a lot easier to test the builds, as you can separate build from platformify (which is why I put it in in first place).

I have tested this by running a doit:full against the Drupal 8 project, with success.


project/templates/drupal.py
```diff
@@ -1,15 +1,16 @@
...
-from .remote import RemoteProject
 import json
 from collections import OrderedDict
 
+from project.baseproject import BaseProject
+from project.remoteproject import RemoteProject
 
-class Drupal7_vanilla(BaseProject):
+
+class Drupal7Vanilla(BaseProject):
```
Crell: Why this rename?  The template repository has an underscore in it, so I fear this would break it.

NickAnderegg:
I like all of this *except* for having to do imports from specific files in the `project` module, e.g. `from project.baseproject import BaseProject`. As there is no overlap in the namespace of the various submodules, it makes a lot more sense to import these classes into `__init__.py` to allow them to be imported from there. For example:

```
from . import baseproject, remoteproject

BaseProject = baseproject.BaseProject
RemoteProject = remoteproject.RemoteProject

__all__ = (
    BaseProject,
    RemoteProject,
)
```

Then imports can be as simple as `from project import BaseProject, RemoteProject`. Thoughts, @xtfer?

project/templates/drupal.py
xtfer: PEP naming convention only (CamelCase class names). 

I’ll try running a build ...

xtfer:
AFAIK doing all your imports in __init__.py is frowned upon, and since I was trying to stick to best practice I didn’t change it. That said, I actually think it’s fine and tend to do so in my own projects. Doubly so in this project, where there is no penalty for loading everything. 

I’ll update the PR with that change.

project/templates/drupal.py
xtfer: Seems to work fine, but fails on a patch so hard to say conclusively.

xtfer:
@NickAnderegg Updated.

project/templates/drupal.py
NickAnderegg: I think this is one PEP 8 convention we can and _should_ violate. I've reverted this change (although it inexplicably doesn't seem to be showing up in the changes I just pushed).

PR #145, opened 2019-10-03
Clean up wp install. Eliminate 2nd install dir.

chadwcarlson:
This is a general cleanup of the Wordpress template.

- There is a single, default, install directory: `wordpress`.
- Removes composer.json/lock from files in favor of upstream from `johnbloch/wordpress`.
- Modifies wordpress's `platformify` action to modify upstream `composer.json`, so that default install location (`wordpress`) is used for defining `extra.installer-paths`.
- Modifies wordpress's `platformify` action to move Platform.sh-specific `wp-config.php` into `wordpress` using `post-install-cmd`.
- Updates README to remove mention of `.platform.template.yaml`.
- Removes composer.json/lock from files in favor of upstream.
- All placeholder files/READMEs previously in `web` have been deleted. 

https://github.com/platformsh-templates/wordpress/pull/3


templates/wordpress/files/composer.json
```diff
@@ -31,20 +31,18 @@
         "sort-packages": true
     },
     "scripts": {
-      "update-packages": [
-          "cp -rf wp/* web/",
-          "rm -rf wp"
+      "movewpcore-todocroot": [
+          "cp -rf wordpress/* web/ && rm -rf wordpress"
```
Crell: Shouldn't this be just a "mv" command...?

chadwcarlson: @Crell `mv wordpress/* web` fails because our template upstream contains `wp-content/`.

```
> mv wordpress/* web
mv: rename wordpress/wp-content to web/wp-content: Directory not empty
```

Which has me thinking... is there a reason we need anything other than `wp-config.php` in `web` to begin with, if all we're doing after install is copying everything over? It looks like it doesn't affect your ability to download plugins/themes. All that's in there now it READMEs, and those subdirectories are a apart of the upstream anyways.

PR #171, opened 2019-11-26
Add Elastic APM with Kibana template.

Crell:
cf: https://github.com/platformsh-templates/elastic-apm/pull/1


templates/elastic-apm/files/kibana/.platform.app.yaml
```diff
@@ -0,0 +1,62 @@
...
+relationships:
+     elasticsearch: "elasticsearch:elasticsearch"
+
+# The size in megabytes of persistent disk space to reserve as part of this application.
+disk: 1024
+
+## Each mount is a pairint of the local path on the application container to
+## the persistent mount where it lives. At this time, only 'shared:files' is
+## a supported mount.
+mounts:
```
chadwcarlson: Isn't this failing because of this mount syntax here?

```
mounts:
     "/run":
         source: local
         source_path: "run"
```

Crell: I guess `source: local` is the default?

PR #172, opened 2019-11-29
Fixes link to patch file in README.

chadwcarlson:
No matching template PR yet, build giving me troubles for some reason.


Crell:
Magento build issues are a known thing, so merging and it will get picked up when the Magento build works again.

PR #179, opened 2019-12-18
Adds strapi template.

chadwcarlson:
https://github.com/platformsh-templates/strapi/pull/1


templates/strapi/files/README.md
```diff
@@ -0,0 +1,63 @@
+# Strapi template for Platform.sh
+
+<p align="center">
+<a href="https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/strapi/.platform.template.yaml&utm_content=strapi&utm_source=github&utm_medium=button&utm_campaign=deploy_on_platform">
+    <img src="https://platform.sh/images/deploy/lg-blue.svg" alt="Deploy on Platform.sh" width="180px" />
+</a>
+</p>
+
+This template builds a Strapi backend for Platform.sh.
```
Crell: Worth noting here that it does not include a front-end, so it's going to not look like much after install.

Crell: (And in the template.yaml file.)

templates/strapi/files/README.md
```diff
@@ -0,0 +1,63 @@
...
+2. Once you have registered the admin user, you will have access to the **Admin Panel**, and from there you can begin adding Content Types to build out the API. By default, no Content Types have been included in the installation *and* any Content Types you add will have secured API endpoints that will make them publicly inaccessible. You will need to update the permissions for each Content Type from the *Users & Permissions* section of the **Admin Panel** after you have created them. If you are unfamiliar with how to create Content Types or modify permissions with Strapi, visit the [Quick Start Guide](https://strapi.io/documentation/3.0.0-beta.x/getting-started/quick-start.html) for detailed instructions.
+3. This template provides only the API for a full project, but you can modify the repository into a [multi-app project](https://docs.platform.sh/configuration/app/multi-app.html#multiple-applications) where Strapi acts as a separate backend application for the front end of your choice. For example, you can repackage this template into a subdirectory called `backend` to create a project structure that looks like this:
```
Crell: If including a multi-app front-end is the common approach... would it make sense to try and get strapi into a subdirectory to begin with, to make a multi-app setup a little easier?  Less file moving for the user that way, but without another directory it will still deploy single-app.

I'm not sure of this myself, but it's worth considering.

templates/strapi/files/config/environments/production/database.js
```diff
@@ -0,0 +1,36 @@
+const config = require("platformsh-config").config();
```
Crell: Does Strapi not have an "All environments" directory for config the way Symfony and Laravel do?

templates/strapi/files/extensions/users-permissions/config/jwt.json
```diff
@@ -0,0 +1,3 @@
+{
+  "jwtSecret": "1dcd9d8e-9e0b-4f84-bada-6683b1cf177d"
```
Crell: This seems like something we shouldn't be committing.  Can it be built off of the PLATFORM_ENTROPY env var?

templates/strapi/files/config/environments/production/request.json
```diff
@@ -0,0 +1,23 @@
+{
+  "session": {
+    "enabled": true,
+    "client": "cookie",
+    "key": "strapi.sid",
+    "prefix": "strapi:sess:",
+    "secretKeys": ["mySecretKey1", "mySecretKey2"],
```
Crell: If these are something the user needs to change, we should probably call that out in the README.

templates/strapi/files/README.md
```diff
@@ -0,0 +1,63 @@
...
+* The `config/(development|production|staging)/server.json` has been modified to read the `port` configuration from the `PORT` Platform.sh environment variable.
+* The start command is set up to run Strapi as a [`development` environment](https://strapi.io/documentation/3.0.0-beta.x/cli/CLI.html#strapi-develop-dev), allowing you to create new Content Types once it is deployed, even on the `master`/production branch of your project. It calls an added `start.sh` script to run `yarn develop`. When taking your application live, it is recommended that you set Strapi for [`production`](https://strapi.io/documentation/3.0.0-beta.x/cli/CLI.html#strapi-start), where Strapi runs without restarts and the creation of new Content Types is not allowed. `start.sh` includes an example environment-specific start command that can be used to accomplish this using the `PLATFORM_BRANCH` Platform.sh environment variable.
```
Crell: I didn't really follow this bullet until I read the script.  I think it should be turned around: The start command calls start.sh, which is configured to run dev mode. We recommend adjusting it to automatically run in prod mode on the master branch; see the script for code to enable.  (Or something like that.)

templates/strapi/files/.platform.app.yaml
```diff
@@ -0,0 +1,45 @@
+# This file describes an application. You can have multiple applications
```
Crell: There's no cron section in here, but Strapi's configuration files talk about cron.  Does it have its own internal cron logic that ignores system cron?  (Seriously people???)

If so, that may be worth documenting, um, somewhere.  Either here or the README, not sure where.

Crell:
It seems an awful lot of this is just a stock upstream starter kit, yes?  If so, shouldn't that be pulled in as a remote and them platformified so that we can stay up to date with upstream improvements to the default files?

templates/strapi/files/backend/.platform.app.yaml
```diff
@@ -0,0 +1,54 @@
...
+type: nodejs:12
+
+dependencies:
+  nodejs:
+    yarn: "1.19.2"
+
+# The hooks that will be triggered when the package is deployed.
+hooks:
+    # Build hooks can modify the application files on disk but not access any services like databases.
+    build: !include
```
Crell: Nice, actually using that syntax style. :smile:

Crell:
The "create first admin" button links to localhost, hard coded, rather than to the current site.

chadwcarlson:
@Crell this is an upstream open issue. TB patch?

templates/strapi/files/README.md
```diff
@@ -0,0 +1,61 @@
...
+  │   ├── extensions
+  │   ├── package.json
+  │   ├── public
+  │   └── yarn.lock
+  └── frontend
+      ├── .platform.app.yaml
+      ├── index.js
+      ├── package-lock.json
+      └── package.json
+  ```
```
Crell: I'm unclear from context how this is different.  Does Strapi include its own front and backend, so it's just a matter of adding a second app.yaml?  Or is the frontend here "some other application"?  It's not clear what this is for.

chadwcarlson: I should update the README to show it's not necessary to move into `backend` since  it's done already now.

But yes, frontend would be a separate app with Vue, Express, etc.

templates/strapi/files/backend/build.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Create quickstart Strapi project.
+yarn create strapi-app tmp-app --quickstart --no-run
```
Crell: Hrm.  So... I'm torn here.

On the one hand, doing a project create in the build hook and copying files around seems like it's going to result in a project that works for first install and not much else.  Certainly it would be hard to run locally, wouldn't it?  

On the other, the alternative seems to be to commit a whole lot of code to the repo.  Which... to be fair, is what we're doing for most other projects that need a "create-project" step.  (Mostly PHP apps.)

I'm not sure which approach is less bad.

chadwcarlson: > On the one hand, doing a project create in the build hook and copying files around seems like it's going to result in a project that works for first install and not much else. Certainly it would be hard to run locally, wouldn't it?

I think it will still work fine regardless of first install, but yet running locally would require you run `platform build` or an equivalent npm script that does this.

> On the other, the alternative seems to be to commit a whole lot of code to the repo. Which... to be fair, is what we're doing for most other projects that need a "create-project" step. (Mostly PHP apps.)

I can always go back to the way I had it before if that makes more sense. In that case I had had updated the areas where the upstream `localhost` linking issue was there.

Crell:
The "create first admin" link is still an issue. I  don't know what "TB patch" means, but we shouldn't ship this without that fixed or the README instructions don't actually work.

PR #204, opened 2020-02-17
Keycloack

otaviojava:
Ref: https://github.com/platformsh-templates/keycloak/pull/1


templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+type: 'java:11'
+
+variables:
+    env:
+        KC_VERSION: 8.0.2
+        JBOSS_HOME: /app
+        JDBC_MARIADB_VERSION: 2.2.3
+        LAUNCH_JBOSS_IN_BACKGROUND: 1
+        #DB_VENDOR: mariadb
```
Crell: Should we remove this commented line?

templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+            cp -f tools/logging.properties standalone/configuration/logging.properties
+            bin/add-user-keycloak.sh -r master --user "admin" --password "password"
+            touch tmp/INSTALLED
+        fi;
+        export DB_PORT=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].port"`
+        export DB_ADDR=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].host"`
+        export DB_PASSWORD=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].password"`
+        export DB_USER=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].username"`
+        export DB_DATABASE=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].path"`
+        export FRONTEND_URL=`echo $PLATFORM_ROUTES|base64 -d|json_pp|./jq -r 'to_entries | map_values(.value + { slug: .key }) | .[] | select(.primary == true) | .slug'`
```
Crell: Wait, do env vars set here actually still apply to the running application?  That... might simplify some things...  Or is that only for Java, not the script languages?

templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+    
+    deploy: |
+        if [ -f tmp/INSTALLED ]; then
+            cp standalone/cfg.bkp/* standalone/configuration/
+            cp -f tools/standalone.xml standalone/configuration/standalone.xml
+            cp -f tools/logging.properties standalone/configuration/logging.properties
+            bin/add-user-keycloak.sh -r master --user "admin" --password "password"
+            touch tmp/INSTALLED
+        fi;
+        export DB_PORT=`echo $PLATFORM_RELATIONSHIPS|base64 -d|json_pp|./jq -r ".database[0].port"`
```
Crell: The json_pp step isn't necessary.  That just does whitespace reformatting, which isn't needed here.

Also, jq doesn't need to be called from the current directory if it's globally installed.

Also, let's put some spaces in the command to make it easier to read.  To wit:

```suggestion
        export DB_PORT=`echo $PLATFORM_RELATIONSHIPS | base64 -d | jq -r ".database[0].port"`
```

templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+        rm -rf standalone/tmp/auth
+        rm -rf domain/tmp/auth
+    
+    deploy: |
+        if [ -f tmp/INSTALLED ]; then
+            cp standalone/cfg.bkp/* standalone/configuration/
+            cp -f tools/standalone.xml standalone/configuration/standalone.xml
+            cp -f tools/logging.properties standalone/configuration/logging.properties
+            bin/add-user-keycloak.sh -r master --user "admin" --password "password"
+            touch tmp/INSTALLED
```
Crell: Is capitalized INSTALLED typical for other Java applications?  We're not capitalizing the touch file for any other templates.

We really should standardize how these touch files are named; we're using them on a number of templates now so consistency would be good.

templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+        KC_VERSION: 8.0.2
+        JBOSS_HOME: /app
+        JDBC_MARIADB_VERSION: 2.2.3
+        LAUNCH_JBOSS_IN_BACKGROUND: 1
+        #DB_VENDOR: mariadb
+hooks:
+    build: |
+        set -e
+
+        if [ ! -d "$PLATFORM_CACHE_DIR"/keycloak-$KC_VERSION ]; then
```
Crell: 1. This build hook looks complex enough it would be easier to read if moved to a separate shell script.
2. It's downloading several things, which means it would be better to leverage the build cache to avoid doing so repeatedly on every build.  That would likely speed up the build considerably.  Let's do that.

templates/keycloak/files/.platform.app.yaml
```diff
@@ -0,0 +1,83 @@
...
+        source: local
+        source_path: log
+    'standalone/deployments':
+        source: local
+        source_path: deployments
+    'standalone/tmp':
+        source: local
+        source_path: tmp
+
+disk: 1024
```
Crell: Total nitpick: I prefer to put the `disk` key before `mounts`.  I think that's more common in the other templates.

templates/keycloak/files/.platform/services.yaml
```diff
@@ -0,0 +1,12 @@
...
+    type: mariadb:10.4
+    disk: 1024
+    configuration:
+        schemas:
+            - main
+        endpoints:
+            mariadb:
+                default_schema: main
+                privileges:
+                    main: admin
```
Crell: This is the default behavior already, so why is it being spelled out?  If it's just to force a password to be generated, let's put that in a comment.

templates/keycloak/files/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+<p align="center">
+<a href="https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/keycloak/.platform.template.yaml&utm_content=keycloak&utm_source=github&utm_medium=button&utm_campaign=deploy_on_platform">
+    <img src="https://platform.sh/images/deploy/lg-blue.svg" alt="Deploy on Platform.sh" width="180px" />
+</a>
+</p>
+
+This project provides a starter kit for Keycloak projects hosted on Platform.sh.
+
+Keycloak is an open source software product to allow single sign-on with Identity Management and Access Management aimed at modern applications and services.
```
Crell:
```suggestion
Keycloak is an Open Source software tool to allow single sign-on with Identity Management and Access Management aimed at modern applications and services.
```

templates/keycloak/files/.platform.app.yaml
otaviojava: Yes, thank you

templates/keycloak/files/.platform.app.yaml
otaviojava: It is not Java, but to this Application and how it was designed.

templates/keycloak/files/.platform.app.yaml
otaviojava: That is terrible, I agree, that is to this application and its design.

templates/keycloak/files/.platform/services.yaml
otaviojava: It is to force the password.

otaviojava: Done.

templates/keycloak/files/.platform.app.yaml
otaviojava: @Crell it has 21 lines, I would like to keep on this file for while, also the condition is to avoid download repeatedly

Crell: ... I am blind and moving too fast these days. :frowning: You're right, it's already using the cache, on this very line.

templates/keycloak/files/.platform.app.yaml
Crell: Oh, this is building off of an INSTALLED file that the app itself already uses?  If so, that's worth a comment.

templates/keycloak/files/.platform.app.yaml
Crell: I mean more for Platform.  Viz, if we were to set env vars in the deploy hook on a PHP app, would that actually get picked up?

templates/keycloak/files/.platform/services.yaml
```diff
@@ -0,0 +1,19 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+
+
+database:
+    type: mariadb:10.4
+    disk: 1024
+#it's just to force a password to be generated, let's put that in a comment.
```
chadwcarlson: delete or rewrite with explanation?

otaviojava: Fixed, thank you

PR #211, opened 2020-02-19
make this idempotent

OriPekelman:


spk:
This is useful when running `doit platformify:rails` multiple time :pray:

PR #219, opened 2020-03-11
Adds phoenix template

chadwcarlson:
https://github.com/platformsh-templates/phoenix_elixir/pull/1


project/phoenix.py
```diff
@@ -0,0 +1,19 @@
+from . import BaseProject, TEMPLATEDIR
+import os
+
+
+class Phoenix_elixir(BaseProject):
+    '''Phoenix_elixir object.
+
+    This is necessary because elixir has some very specific naming conventions, that prevent the template from being called
```
Crell: FFS, Elixir!  Have you learned nothing from Go?

project/phoenix.py
```diff
@@ -0,0 +1,19 @@
...
+    only underscore.
+
+    This constructor sets the 'github_name' attribute to get around the naming convention checks in the BaseProject constructor.
+
+    Here, it simply reuses 'name' to match the directory name exactly.
+    '''
+    def __init__(self, name):
+        self.name = name
+        self.github_name = name
+        self.builddir = os.path.join(TEMPLATEDIR, self.name, 'build/')
```
Crell: Shouldn't this call super rather than repeating the parent constructor's code?

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+# See https://docs.platform.sh/user_guide/reference/platform-app-yaml.html
+
+# The name of this app. Must be unique within a project.
+name: app
+
+# The runtime the application uses.
+type: elixir:1.9
+
+build:
+    flavor: none
```
Crell: Is there a flavor to override?  If we have to override it on even our first template, it means we should push back on Engineering and remove flavors entirely.  (Which is what we had planned to do in the first place for non-PHP, but...)

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+variables:
+    env:
+        MIX_ENV: 'prod'
+        SECRET_KEY_BASE: $PLATFORM_PROJECT_ENTROPY
```
Crell: Wait... that works?

templates/phoenix_elixir/files/.platform.app.yaml
chadwcarlson: @Crell this  is what I've found the flavor assumes so far  https://pr-1198-pj4okaq-652soceglkw4u.eu-3.platformsh.site/languages/elixir.html#assumptions

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+        mix do deps.get --only prod, deps.compile, compile
+        if grep -q phoenix mix.lock
+        then
+            npm install --prefix ./assets
+            npm run deploy --prefix ./assets
+            mix phx.digest
+        fi
+    deploy: |
+        mix do ecto.setup
+relationships:
```
Crell:
```suggestion

relationships:
```

templates/phoenix_elixir/files/.platform.app.yaml
chadwcarlson: It's removed on phoenix because it breaks the compile step, but in the "vanilla" elixir template I'll push in a moment, the flavor works fine.

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+    database: "database:postgresql"
+# The configuration of app when it is exposed to the web.
```
Crell:
```suggestion

# The configuration of app when it is exposed to the web.
```

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+    locations:
+        /:
+            # Route all requests to the Lisp app, unconditionally.
+            # If you want some files served directly by the web server without hitting Elixir, see
+            # https://docs.platform.sh/configuration/app/web.html
+            allow: false
+            passthru: true
+
+# The size of the persistent disk of the application (in MB).
+disk: 512
```
Crell: No mounts?  If there's no mounts at all disk can be set to the minimum of 128 to minimize wasted space.

templates/phoenix_elixir/files/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+<p align="center">
+<a href="https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/phoenix_elixir/.platform.template.yaml&utm_content=phoenix_elixir&utm_source=github&utm_medium=button&utm_campaign=deploy_on_platform">
+    <img src="https://platform.sh/images/deploy/lg-blue.svg" alt="Deploy on Platform.sh" width="180px" />
+</a>
+</p>
+
+This template builds a simple application using Phoenix.
+
+Phoenix is a web framework written in Elixir that implements the server-side MVC pattern, prioritizing speed and concurrency.
```
Crell: That may be how they describe themselves, but there is no such thing as server-side MVC.  It doesn't exist. Don't perpetuate that lie...

templates/phoenix_elixir/files/config/prod.secret.exs
```diff
@@ -0,0 +1,26 @@
+# In this file, we load production configuration and secrets
+# from environment variables. You can also hardcode secrets,
+# although such is generally not recommended and you have to
+# remember to add this file to your .gitignore.
+use Mix.Config
+
+secret_key_base =
+  System.get_env("PLATFORM_PROJECT_ENTROPY") ||
```
Crell: This is reading from the original Platform variable, but in .platform.app.yaml it's being mapped over to the default variable name Elixir uses.  Why do both?

Crell: To support local dev, reading from the default env var but mapping the Psh value to it seems like the better option.  Viz, have this line read from the original default variable name.

templates/phoenix_elixir/files/lib/template_elixir_web/endpoint.ex
```diff
@@ -0,0 +1,46 @@
...
+  plug Plug.MethodOverride
+  plug Plug.Head
+
+  # The session will be stored in the cookie and signed,
+  # this means its contents can be read but not tampered with.
+  # Set :encryption_salt if you would also like to encrypt it.
+  plug Plug.Session,
+    store: :cookie,
+    key: "_template_elixir_key",
+    signing_salt: "TaQ2DjZK"
```
Crell: That seems like a value that shouldn't be in Git.

project/phoenix.py
chadwcarlson: no way jose

templates/phoenix_elixir/files/lib/template_elixir/repo.ex
```diff
@@ -0,0 +1,13 @@
+defmodule TemplateElixir.Repo do
+  use Ecto.Repo,
+    otp_app: :template_elixir,
+    adapter: Ecto.Adapters.Postgres
+	
+    @doc """
+    Dynamically loads the repository configuration from the environment variables.
+    """
+    def init(_) do
+      url = System.get_env("DATABASE_URL")
```
Crell: It doesn't seem this env var gets set from anywhere?  Or did I miss it?

project/phoenix.py
chadwcarlson:
```
    def __init__(self, name):
        self.github_name = name
        return super(Phoenix_elixir, self).__init__(name)
```

templates/phoenix_elixir/files/config/config.exs
```diff
@@ -0,0 +1,30 @@
...
+# General application configuration
+use Mix.Config
+
+config :template_elixir,
+  ecto_repos: [TemplateElixir.Repo]
+
+# Configures the endpoint
+config :template_elixir, TemplateElixirWeb.Endpoint,
+  url: [host: "localhost"],
+  secret_key_base:   System.get_env("PLATFORM_PROJECT_ENTROPY"),
```
Crell: Huh.  Also redundant?  Why does this get set 3 times?

templates/phoenix_elixir/files/config/prod.secret.exs
chadwcarlson: 👍

templates/phoenix_elixir/files/config/config.exs
chadwcarlson: I think actually that in the end
- set in app.yaml
- read in config.exs

is sufficient

templates/phoenix_elixir/files/.platform.app.yaml
Crell: We should still be pushing back on Engineering.  We decided after Node that we didn't want to use flavors anymore as it was too much magic, so I don't know why that's there at all.

chadwcarlson: That's fair, I think Ori's for removing it entirely.

chadwcarlson: So I guess then I'll remove flavor mentions from the simple template too for now.

PR #221, opened 2020-03-18
Adds mattermost.

chadwcarlson:
https://github.com/platformsh-templates/mattermost/pull/1


templates/mattermost/.platform.template.yaml
```diff
@@ -0,0 +1,107 @@
+# Platform.sh Project Initialization Template
```
Crell: The first half of this file appears twice.

templates/mattermost/files/.platform.app.yaml
```diff
@@ -0,0 +1,64 @@
...
+    source_path: files/log-files
+  'data':
+    source: local
+    source_path: files/data-files
+  'dist/files':
+    source: local
+    source_path: files/dist-files
+  'plugins':
+    source: local
+    source_path: files/plugins
```
Crell: The files prefix on all of these is unnecessary.

PR #237, opened 2020-04-13
Drupal 8 updates and Redis

Crell:
* Updates to Drupal 8.8.5, which fixes the installer issues.
* It also fixes the bug that kept Redis from working if enabled by default, so now Redis is enabled by default.
* Redis is custom-compiled.
* Drupal 8 now runs on PHP 7.4.

Same changes applied to Multisite.

And the same changes to Opigno, except for the PHP upgrade; some of its dependent modules still have syntax errors on PHP 7.3 and later.

cf: 

* https://github.com/platformsh-templates/drupal8/pull/9
* https://github.com/platformsh-templates/drupal8-multisite/pull/8
* https://github.com/platformsh-templates/drupal8-opigno/pull/4


templates/drupal8-multisite/.platform.template.yaml
```diff
@@ -39,7 +39,7 @@ info:
   # Each note object is displayed as a small section heading with content below. Supports limited HTML.
   notes:
     - heading: "Apps & Services"
-      content: "PHP 7.2<br/>MariaDB 10.2<br />Redis 5.0"
+      content: "PHP 7.4<br/>MariaDB 10.2<br />Redis 5.0"
```
chadwcarlson, line 42:
```suggestion
      content: "PHP 7.4<br/>MariaDB 10.4<br />Redis 5.0"
```

PR #250, opened 2020-05-06
Phoenix

chadwcarlson:
https://github.com/platformsh-templates/phoenix_elixir/pull/2

Resubmits https://github.com/platformsh/template-builder/pull/219


project/phoenix.py
```diff
@@ -0,0 +1,16 @@
+from . import BaseProject, TEMPLATEDIR
+import os
+
+
+class Phoenix_elixir(BaseProject):
+    '''Phoenix_elixir object.
+    This is necessary because elixir has some very specific naming conventions, that prevent the template from being called
```
Crell, line 7:
```suggestion
    This is necessary because elixir has some very specific naming conventions that prevent the template from being called
```

project/phoenix.py
```diff
@@ -0,0 +1,16 @@
...
+        simply 'phoenix' (since it is trying to compile the phoenix Hex module), and the project cannot contain the '-' character,
+    only underscore.
+    This constructor sets the 'github_name' attribute to get around the naming convention checks in the BaseProject constructor.
+    Here, it simply reuses 'name' to match the directory name exactly.
```
Crell, line 11: This level of documentation: <3

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+# The runtime the application uses.
+type: elixir:1.9
+
+variables:
+    env:
+        SECRET_KEY_BASE: $PLATFORM_PROJECT_ENTROPY
+        MIX_ENV: 'prod'
+
+# This is assuming we have nothing in the build flavour
```
Crell, line 17: Can this comment be removed now?

chadwcarlson: yessir. updated

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+# The configuration of app when it is exposed to the web.
+web:
+    commands:
+        start: mix do platformsh.config, phx.server --no-halt --no-compile
+    locations:
+        /:
+            # Route all requests to the app, unconditionally.
+            allow: false
+            passthru: true
```
Crell, line 42: Is this typical phoenix configuration?  No static files at all?

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+# The size of the persistent disk of the application (in MB).
+disk: 128
```
Crell, line 45: If there's no mounts, `disk` is now optional as of Git 4.

chadwcarlson: removed

templates/phoenix_elixir/files/config/config.exs
```diff
@@ -0,0 +1,30 @@
...
+
+# General application configuration
+use Mix.Config
+
+config :template_elixir,
+  ecto_repos: [TemplateElixir.Repo]
+
+# Configures the endpoint
+config :template_elixir, TemplateElixirWeb.Endpoint,
+  url: [host: "localhost"],
```
Crell, line 15: Eh?

chadwcarlson: generated by the ` mix phx.new hello` command for creating a new app.

templates/phoenix_elixir/files/.platform.app.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+# The name of this app. Must be unique within a project.
+name: app
+
+# The runtime the application uses.
+type: elixir:1.9
+
+variables:
+    env:
+        SECRET_KEY_BASE: $PLATFORM_PROJECT_ENTROPY
```
spk, line 14: I had issue with last phoenix release I was force to have a 64 chars length here and I think PLATFORM_PROJECT_ENTROPY is 56 chars currently, should we let the user change it or let it fail ?

spk (commented):
Too bad I did not see this PR before doing my project :pray:

templates/phoenix_elixir/files/.platform.app.yaml
Crell: Laravel has a similar problem.  What we do there (albeit in PHP space) is compute a proper-length hash using the entropy value as a seed.

cf: https://github.com/platformsh/laravel-bridge/blob/master/platformsh-laravel-env.php#L30

Might something similar-ish work here?

spk: I don't know that much Laravel but yes look like we could do something like this, don't know how but its interesting topic

PR #255, opened 2020-05-20
Discourse

Crell:
Trying to port https://github.com/OriPekelman/discourse-template into template-builder.

Current problems:

1) [x] WSOD on missing domain name, even if the domain name is set in an env var as in the example above.
2) [x] Setting the domain name in an env var is a hard-no, as it breaks as soon as you start branching things.  That still has to get fixed.
3) [ ] I... don't know how anyone is expected to update their own installs once this is deployed.
4) [x]  There's still lots of errors in the build that don't *seem* like they're actually breaking anything, but it would be good to clean up if we can.
5) [ ] README file still needs to be ported.
6) [ ] Needs logo.


templates/discourse/files/.platform/applications.yaml
```diff
@@ -0,0 +1,143 @@
...
+          # install nvm
+
+          unset NPM_CONFIG_PREFIX
+          curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.35.2/install.sh | dash
+          export NVM_DIR="$PLATFORM_APP_DIR/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm current
+          nvm install 12.16.3
+          # Install globally some more required npm packages that are not in package.json
+          npm i -g "yarn" "svgo" "gifsicle@4.0.1" "uglify@<3"
```
Crell, line 29: ... Why aren't they?  And why not then use dependencies?

templates/discourse/files/.platform/applications.yaml
```diff
@@ -0,0 +1,143 @@
...
+          # create symbolic link form app/tmp to /tmp so Bootsnap doesn't die.
+          ln -s /tmp tmp
+          # install nvm
+
+          unset NPM_CONFIG_PREFIX
+          curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.35.2/install.sh | dash
+          export NVM_DIR="$PLATFORM_APP_DIR/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm current
+          nvm install 12.16.3
```
Crell, line 27: @todo Regeneralize this.

templates/discourse/files/.platform/applications.yaml
```diff
@@ -0,0 +1,143 @@
...
+          # Install Gems
+          bundle install --retry 3 --jobs 4
+          bundle exec rake plugin:install_all_gems
+          # Optionally install the CLI
+          # curl -sS https://platform.sh/cli/installer | php
+      deploy: |
+          unset NPM_CONFIG_PREFIX
+          export NVM_DIR="$PLATFORM_APP_DIR/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm use 12.16.3
```
Crell, line 55: TODO: Re-generalize this based on the variable above.

templates/discourse/files/config/discourse.platform.sh.conf
```diff
@@ -0,0 +1,26 @@
...
+smtp_authentication = none
+# enable TLS encryption for smtp connections
+smtp_enable_start_tls = false
+serve_static_assets = true
+config.assets.compile = true
+config.assets.digest = false
+
+# you should export in the environment using the CLI or set here the specifics
+# platform vset -emaster env:DISCOURSE_DEVELOPER_EMAILS myemail@example.com
+# platform vset env:DISCOURSE_MAXMIND_LICENSE_KEY get_a_key_from_maxmind
```
Crell, line 21: These don't seem to do anything...?  The site installs without them.

templates/discourse/files/config/discourse.platform.sh.conf
```diff
@@ -0,0 +1,26 @@
...
+# platform vset -emaster env:DISCOURSE_HOSTNAME $(platform route:get --primary -P url| cut -d'/' -f3)
+
+#developer_emails=myemail@example.com
+# FIXME this should work through the environment variable and doesn't?
+hostname = <%= ENV['DISCOURSE_HOSTNAME']%>
```
Crell, line 26: It doesn't?  Do any of the other env vars here?  I've not been able to verify.

Crell: If it doesn't work, then why is this env var being set manually, per the instructions?  This doesn't compute.

templates/discourse/files/.platform/applications.yaml
OriPekelman: I'd imagine they don't install with our antiquated base node installation. You could test .. but that would be my assumption. Otherwise I would have done it as a dependency.

templates/discourse/files/.platform/applications.yaml
OriPekelman: It is better to use "nvm install stable" or something like this. Look at my playbook repo for a much better take on installing NVM (that enables caching the nvm downloads).

templates/discourse/files/config/discourse.platform.sh.conf
OriPekelman: You will very probably have errors if you don't. Possibly they corrected the install process around those two .. but I don't believe so. At any rate it is good to have the "developer email" set to the project admin.

templates/discourse/files/config/discourse.platform.sh.conf
OriPekelman: It seems discourse overrides the Env variables from the config file (which is really a weird take on configuration precedence. I don't want us to change any of the files in the repo if we can avoid it. So we are weirdly creating this one, which just makes sure the order of precedence is correct.

templates/discourse/files/.platform/applications.yaml
Crell: Wouldn't `stable` mean it changes when upstream changes what it considers stable?

templates/discourse/.platform.template.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Templates are a small amount of information supporting a template URL. Each template is selectable at the project-creation step.
+info:
+  # Unique machine name, prefaced by a vendor or organization identifier
+  id: platformsh/discourse
+  # The human-readable name of the template.
+  name: Discourse
+  # Human-readable descriptive text for the template. Supports limited HTML.
+  description: |
+      <p>This template builds Discourse on Platform.sh.  Note that it does require some post-install setup.</p>
+      <p>Discourse is a community forum application build in Ruby.</p>
```
chadwcarlson, line 28:
```suggestion
      <p>Discourse is a community forum application built in Ruby.</p>
```

templates/discourse/files/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+<p align="center">
+<a href="https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/discourse/.platform.template.yaml&utm_content=discourse&utm_source=github&utm_medium=button&utm_campaign=deploy_on_platform">
+    <img src="https://platform.sh/images/deploy/lg-blue.svg" alt="Deploy on Platform.sh" width="180px" />
+</a>
+</p>
+
+This template builds Discourse on Platform.sh.  Note that it does require some post-install setup.
+
+Discourse is a community forum application build in Ruby.
```
chadwcarlson, line 11:
```suggestion
Discourse is a community forum application built in Ruby.
```

templates/discourse/files/deploy.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+unset NPM_CONFIG_PREFIX
```
chadwcarlson, line 3: You're running this in the build and the deploy?

Crell: Yes, both build and runtime need to be shifted over into modern-node mode.  That's what our docs say to do.

PR #280, opened 2020-07-02
Automated process and basic testing capabilities

moisesbenzan:
With is pull request, template-builder will be able to:

a) automatically updating of templates with the whole (pull, update, platformify, push, pull request, test, merge) lifecycle
b) Ability to add multiple python tests to a template via a function with the following signature prototype: `def test_1(url: str) -> Boolean:`

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -129,6 +129,13 @@ def task_pull_request():
         yield {
             'name': project.name,
             'actions': [project.pull_request],
+            'params': [{
+                'name': 'token',
+                'short': 't',
+                'long': 'token',
+                'default': None,
+                'help': 'Parses the github token from the commandline, overriding the GITHUB_TOKEN env var.'
+                }]
         }
 
 def task_push():
@@ -169,6 +176,13 @@ def task_test():
         yield {
             'name': project.name,
             'actions': [project.test],
+            'params': [{
+                'name': 'token',
+                'short': 't',
+                'long': 'token',
+                'default': None,
+                'help': 'Parses the github token from the commandline, overriding the GITHUB_TOKEN env var.'
+                }]
         }
 
 def task_merge():
@@ -181,6 +195,13 @@ def task_merge():
         yield {
             'name': project.name,
             'actions': [project.merge_pull_request],
+            'params': [{
+                'name': 'token',
+                'short': 't',
+                'long': 'token',
+                'default': None,
+                'help': 'Parses the github token from the commandline, overriding the GITHUB_TOKEN env var.'
+                }]
         }
 
 def task_full():
@@ -196,4 +217,11 @@ def task_full():
                          for action in ['cleanup', 'init', 'update', 'platformify', 'branch', 'push', 'pull_request', 'test', 'merge']
                          ],
             'actions': [],
+            'params': [{
+                'name': 'token',
+                'short': 't',
+                'long': 'token',
+                'default': None,
+                'help': 'Parses the github token from the commandline, overriding the GITHUB_TOKEN env var.'
+                }]
         }
--- a/dodo.py
+++ b/dodo.py
@@ -119,6 +119,18 @@ def task_branch():
         }
 
 
+def task_pull_request():
+    """
+    DoIt Task: Creates a pull request from the update branch to master
+
+    Usage: doit pull_request:<project>
+    """
+    for project in ALL_PROJECTS:
+        yield {
+            'name': project.name,
+            'actions': [project.pull_request],
+        }
+
 def task_push():
     """
     DoIt Task: Pushes a prepared branch to GitHub.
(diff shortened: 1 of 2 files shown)


project/__init__.py
```diff
@@ -98,6 +108,63 @@ def push(self):
...
+
+        results = []
+        print(f"Running {len(self.TEST_FUNCTIONS)} tests.")
+        for test in self.TEST_FUNCTIONS:
+            for url in urls_to_test:
+                results.append(test(url))
+
+        return all(results)
+
+    def merge_pull_request(self):
```
Crell, line 139: I don't think we want any automatic merging of PRs.  PRs should always have human validation.  Create the PR, sure, but humans should be the ones doing all merges (since there is often testing needed that goes beyond what can be automated easily).

project/__init__.py
```diff
@@ -25,10 +26,19 @@ class BaseProject(object):
...
+    
+    '''Amount of seconds to delay test execution after target_url is available.
+    This is due to the fact that some apps take longer to really deploy than what is reported back to the status check.
+    '''
+    TEST_DELAY = 10
 
     def __init__(self, name):
         self.name = name
         self.builddir = os.path.join(TEMPLATEDIR, self.name, 'build/')
+        self.github_token = os.getenv('GITHUB_TOKEN', None) # Maybe add a validation logic to raise an error if not present ?
```
Crell, line 38: Fail gracefully, please. :smile:  I'm not sure what gracefully is in this case, though.

Is there a way to support it as a command line parameter in addition to an env var?  I hate having env references this deep in the code.  It makes testing and development much harder.

project/__init__.py
moisesbenzan: It could be left in the code as a feature to be triggered manually instead of by the full pipeline. Or do we want that feature completely removed?

project/__init__.py
moisesbenzan: For the failing gracefully part, at most we could print to `Cannot execute this action because the token is not set` and fail the action.

A command line parameter could be supported on a per subcommand basis but not on operations that run the whole chain. What I mean is, it can be supported for `test:wordpress` which runs ***only*** the testing portion but not for `full:wordpress` which executes the whole command chain of `cleanup`, `init`, `update`, `platformify`, `branch`, `pull`, `pull_request`, `test`, `merge`.

project/__init__.py
ludoge: The whole point of this exercise is to fully automate the template updating process (running in travis CI daily) and take everything out of human hands. You can't have human interaction in CI. If there are tests that are non-trivial to automate there will be non-trivial implementations to be made. "Every pull request should have human validation" is a valid point regarding code changes that need _review_, but here it's a matter of updating dependencies and no human needs to sift through those. On the engineering side, all vendored packages and container images get automated releases and I don't see how this is any different.

project/__init__.py
ludoge: It's fairly standard (again, this wants to run in CI) to supply authentication tokens as environment variables. Not sure where the "much harder" here comes from, as long as we're failing gracefully and explicitly if the variable is not supplied.

dodo.py
```diff
@@ -143,6 +155,29 @@ def task_rebuild():
...
+    """
+    for project in ALL_PROJECTS:
+        yield {
+            'name': project.name,
+            'actions': [project.test],
+        }
+
+def task_merge():
+    """
+    DoIt Task: Runs smoke tests against branches on GitHub.
```
ludoge, line 172: (this is probably _my_ copy paste error but still) The docstring doesn't seem to match what the command does

project/__init__.py
Crell: I'm completely on board with automation replacing 90% of the process.  But I don't have faith that automated tests will be able to fully verify that a given change doesn't break.  Eg, right now I am, once again, tracking down an intermittent issue with Drupal's installer completing successfully but not resulting in a working site at the end, sometimes.  I don't expect automation to be able to step through an interactive Ajax based installer and then verify that the resulting site works correctly.

Even just updating dependencies can cause something to break.  I would much prefer the merge button always be pressed by a human, even if everything before that is a nightly automated task.

The dependencies underpinning some of these applications are far, far less stable than the ones that underpin the container images.  I wish that weren't true, but...

project/__init__.py
Crell: Having the data come in via an env var is fine; it's having the env var that deep in the code that bothers me.  I would be more comfortable with the env var being mapped to some config value early on (Global in dodo.py? Not sure), and then the task reading from that.  As for failing gracefully, what happens if GITHUB_TOKEN is not set?  If there is a stack trace anywhere in the output, then it's not graceful.  It needs to be checked, caught, and a useful error messages given instead.  

"Die the way None would normally die" is not graceful.

dodo.py
moisesbenzan: Resolved.

project/__init__.py
```diff
@@ -226,3 +234,11 @@ def get_test_urls(self):
         print(f"Delaying execution of tests by {self.TEST_DELAY} seconds.")
         time.sleep(self.TEST_DELAY)
         return urls
+   
+    def set_github_token(self, token):
+       # CMD line token overrides env var token.
+       if token is not None:
+           self.github_token = token
+       # If token is still not set, alert user. Pd: Cannot failt gracefully as it is not on main task code.
```
Crell, line 242: Pd?

moisesbenzan: Postdata

PR #323, opened 2020-09-16
Update start command to use npx next.

chadwcarlson:
https://github.com/platformsh-templates/nextjs/pull/3

### DIFF
--- a/project/nextjs.py
+++ b/project/nextjs.py
@@ -1,10 +1,6 @@
 from . import BaseProject
-import os.path
-from glob import glob
-from . import TEMPLATEDIR
 
 class Nextjs(BaseProject):
     updateCommands = {
         'yarn.lock': 'yarn upgrade'
     }
-    
--- a/project/nextjs.py
+++ b/project/nextjs.py
@@ -1,6 +1,21 @@
 from . import BaseProject
 
 class Nextjs(BaseProject):
+    # The BaseProject `update` task will loop through a list of `updateCommands` for 
+    # a number of different language package managers and run them if their respective
+    # dependency files (i.e. package.json) are present. 
+    # 
+    # Node.js has two popular package mangers - npm and yarn. Both use package.json to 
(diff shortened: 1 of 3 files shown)


project/nextjs.py
```diff
@@ -0,0 +1,6 @@
+from . import BaseProject
+
+class Nextjs(BaseProject):
+    updateCommands = {
+        'yarn.lock': 'yarn upgrade'
+    }
```
Crell, line 6: Why not just add this to the main list in the parent class?  There's nothing Nextjs specific about it.

chadwcarlson: Because package-lock won't ever be used, but it will be created when going through the whole list of `updateCommands` - so it's just an extra file in the repo with no use, pure template-builder artifact. Nextjs recommends using yarn, and running `yarn` with a `yarn.lock` and a `package-lock.json` gives you a warning that it's not recommended to have both lock files present.

chadwcarlson: So unless we introduce some logic into `update` & `package_update_actions()` to only `yarn upgrade` if there's a `yarn.lock` (if present, then skip `npm update`) this seemed like the quickest fix.

project/nextjs.py
```diff
@@ -0,0 +1,21 @@
...
+    # define dependencies, but resolve to different lock files (package-lock.json & yarn.lock).
+    # When deployed, we'll only be using one of them, and it's generally recommended to not have
+    # lock files from both package managers present in the same repo. 
+    # 
+    # Next.js also recommends the use of yarn over npm. 
+    # 
+    # Because `update` loops through `updateCommands`, simply adding an upgrade command for
+    # yarn will result in that exact situation we'd like to avoid. 
+    # 
+    # So, `updateCommands` is redefined here for  Next.js to avoid all of that.
```
Crell, line 18: A comment that long and detailed is... so amazingly on-brand for you. :-)

PR #339, opened 2020-10-06
Update to NC 20, and remove the old custom Redis script.

Crell:
cf: https://github.com/platformsh-templates/nextcloud/pull/12


templates/nextcloud/files/_config/config.php
```diff
@@ -44,7 +44,7 @@
   // FIXME get something more reasonable here
   'logfile' => '/tmp/logs/nextcloud.log',
   'dbtype' => 'mysql',
-  'version' => '18.0.1',
+  'version' => $_ENV['NEXTCLOUD_VERSION'],
```
Crell, line 47: I am not convinced this is the best solution, but it's what I came up with that worked.  I don't know if that's going to cause issues later in future upgrades.

PR #340, opened 2020-10-06
Add default lando config for some symfony based templates

pirog:
@Crell and @pirog to review this

### DIFF
--- a/templates/laravel/files/.lando.upstream.yml
+++ b/templates/laravel/files/.lando.upstream.yml
@@ -1,0 +1,32 @@
+# This file sets some good defaults for local development using this platform.sh
+# template with Lando.
+#
+# Note that you should not edit this file so it can continue to receive upstream
+# updates. If you wish to change the values below then override them in your
+# normal .lando.yml.
+
+# These both allow you to test this template without needing a site on platform.sh
+# However you will want to replace them in your .lando.yml
+name: platformsh-laravel
+recipe: platformsh
+
+config:
+
+  # This section overrides platform.sh configuration with values that make more
+  # sense for local development.
+  #
+  # Note that "app" is the name of the application defined in your
+  # .platform.app.yaml or applications.yaml.
+  overrides:
+    app:
+      variables:
+        env:
+          APP_ENV: development
+          APP_DEBUG: 1
+
+
+# These are tools that are commonly used during development for this template.
+tooling:
+  artisan:
+    cmd: /app/artisan
+    service: app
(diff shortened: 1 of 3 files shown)


templates/symfony5/files/.lando.upstream.yml
```diff
@@ -0,0 +1,32 @@
+# This file sets some good defaults for local development using this platform.sh
```
Crell, line 1:
```suggestion
# This file sets some good defaults for local development using this Platform.sh
```

templates/symfony5/files/.lando.upstream.yml
```diff
@@ -0,0 +1,32 @@
...
+# template with Lando.
+#
+# Note that you should not edit this file so it can continue to receive upstream
+# updates. If you wish to change the values below then override them in your
+# normal .lando.yml.
+
+# These both allow you to test this template without needing a site on platform.sh
```
Crell, line 8:
```suggestion
# These both allow you to test this template without needing a site on Platform.sh
```

templates/symfony5/files/.lando.upstream.yml
```diff
@@ -0,0 +1,32 @@
...
+# However you will want to replace them in your .lando.yml
+name: platformsh-symfony5
+recipe: platformsh
+
+config:
+
+  # This section overrides platform.sh configuration with values that make more
```
Crell, line 15:
```suggestion
  # This section overrides Platform.sh configuration with values that make more
```

templates/laravel/files/.lando.upstream.yml
```diff
@@ -0,0 +1,31 @@
...
+  # This section overrides Platform.sh configuration with values that make more
+  # sense for local development.
+  #
+  # Note that "app" is the name of the application defined in your
+  # .platform.app.yaml or applications.yaml.
+  overrides:
+    app:
+      variables:
+        env:
+          APP_ENV: development
```
pirog, line 24: @Crell researching this a bit more i think i choose this because of the docs here https://github.com/platformsh/laravel-bridge#common-environment-variables-not-set, i took `development` to be a value as written. However, doing more research it looks like this is also often set to `local` (see https://github.com/platformsh-templates/laravel/blob/master/.env.example#L2).

Most important thing seems to be that it is just not set to `production`. I'm happy with either `dev`, `local` `development` or `lando`.

pirog:
@Crell made the changes we discussed and dropped a comment above as well. This also assumes we remove the `install-redis.sh` script. If we do not do that then this template will fail on a `lando rebuild`.

PR #341, opened 2020-10-06
Add default lando config for some drupaly templates

pirog:
@Crell and @pirog to review this


templates/drupal8/files/.lando.upstream.yml
```diff
@@ -0,0 +1,32 @@
...
+    app:
+      variables:
+        d8config:
+          "system.file:path:temporary": "/tmp"
+        d8settings:
+          "skip_permissions_hardening": 1
+
+
+# These are tools that are commonly used during development for this template.
+tooling:
```
Crell, line 30: Should we also add Drupal Console here?  We install it by default on D8; I have no idea how many people actually use it, though.

pirog: We've seen decreasing usage over the last year or so but if its in the template then we should have a tooling route for it as well.

PR #346, opened 2020-10-12
Update gatsby-strapi.

chadwcarlson:
https://github.com/platformsh-templates/gatsby-strapi/pull/9


templates/gatsby-strapi/files/strapi/.platform.app.yaml
```diff
@@ -1,24 +1,35 @@
...
-    # Build hooks can modify the application files on disk but not access any services like databases.
-    build: !include
-      type: string
-      path: build.sh
+    build: |
+        # Download dependencies and build Strapi.
+        yarn
+        yarn build
+
+        # Move extension overrides out of soon to be mounts.
```
Crell, line 28: Grammar parse error.

Crell: I think maybe you mean "soon-to-be mounts"?  It's clumsy wording either way.

templates/gatsby-strapi/files/strapi/.platform.app.yaml
```diff
@@ -30,28 +41,48 @@ relationships:
...
     'extensions':
         source: local
         source_path: extensions
+    # Allow for media uploads at runtime.
+    'public/uploads':
+        source: local
+        source_path: uploads
+
+
+
```
Crell, line 88:
```suggestion

```

templates/gatsby-strapi/files/gatsby/.platform.app.yaml
```diff
@@ -1,16 +1,34 @@
...
 # The hooks that will be triggered when the package is deployed.
 hooks:
     # Build hooks can modify the application files on disk but not access any services like databases.
-    post_deploy: |
-      npm run build
+    build: |
+        yarn
+    # Delay Gatsby's build until the post deploy hook, when Strapi's content is accessible.
+    deploy: |
+        yarn build
```
Crell, line 30: Uh, based on the comment is this supposed to be a post_deploy hook?

project/__init__.py
```diff
@@ -25,6 +25,7 @@ class BaseProject(object):
         'Gemfile': 'bundle update',
         'package.json': 'npm update',
         'go.mod': 'go get -u all',
+        'yarn.lock': 'yarn upgrade'
```
RudyWeber, line 28: Can you explain why this is needed?

As I understand it, we want to update lock files whenever dependencies have changed. For JS projects, these dependencies are declared in `package.json`. So a change in `yarn.lock` means that a change in `package.json` already occurred. This seems to already be handled by `'package.json': 'npm update'` (which is overwritten in projects using Yarn if I understood correctly.) So why do we need this line?
Also, do some projects use npm and others use Yarn? Why don’t we do `'package.json': 'yarn upgrade'` and avoid `npm` entirely? This way we wouldn’t have to overwrite `package.json` in `updateCommands` for some projects and not for others.

Also, unless you use `yarn install --frozen-lockfiles`, Yarn will update `yarn.lock` if needed (i.e. if the information in the lock files aren't enough to satisfy all the dependencies constraints declared in `package.json`. This includes new dependencies, dependencies deletion, change in version numbers.) So `yarn install`ing a project in which a `package.json` file has changed should update the `yarn.lock` file at the same time. That would make the `'pakcage.json'` updateCommand useless I think?

I might lack some background information/implicit knowledge on the project to understand these choices.

templates/gatsby-strapi/files/strapi/config/database.js
```diff
@@ -0,0 +1,62 @@
+const config = require("platformsh-config").config();
+
+let dbRelationship = "postgresdatabase";
```
RudyWeber, line 3: This is never re-assigned so it can be a `const`.

templates/gatsby-strapi/files/gatsby/gatsby-config.js
```diff
@@ -1,7 +1,7 @@
 const config = require("platformsh-config").config();
 
 var backend_route = "";
```
RudyWeber, line 3: `let` should be preferred over `var`.

templates/gatsby-strapi/files/gatsby/gatsby-config.js
```diff
@@ -1,7 +1,7 @@
...
-if ( config.isValidPlatform() ) {
+if ( config.isValidPlatform() && !config.inBuild()) {
```
RudyWeber, line 4: Extra space before `config`.
(I know there were spaces at both sides of the expression before but there's no space for the same code in `templates/gatsby-strapi/files/strapi/config/database.js` L15 so I think it's better to be consistent.)

PR #349, opened 2020-10-13
Add good Lando defaults for the TYPO3 template

pirog:
@Crell and @pirog to review. Specifically need to discuss:

* Any TYPO3 envvars we need to set that are best for the localdevs?
* `lando rebuild` fails because the `build` hook tries to reinstall TYPO3 even though it's already installed? Should we make the install conditional?

This should be considered a draft for now.

### DIFF
--- a/templates/typo3/files/.lando.upstream.yml
+++ b/templates/typo3/files/.lando.upstream.yml
@@ -1,0 +1,18 @@
+# This file sets some good defaults for local development using this Platform.sh
+# template with Lando.
+#
+# Note that you should not edit this file so it can continue to receive upstream
+# updates. If you wish to change the values below then override them in your
+# normal .lando.yml.
+
+# These both allow you to test this template without needing a site on Platform.sh
+# However you will want to replace them in your .lando.yml
+name: platformsh-typo3
+recipe: platformsh
+
+# These are tools that are commonly used during development for this template.
+tooling:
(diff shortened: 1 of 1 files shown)


Crell:
Tagging @chadwcarlson

pirog:
@chadwcarlson, @Crell and i need help on:

* Are there any TYPO3 specific envvars or config that are typically set for local development? If so, what are they and what should we set them to?
* Do you have any suggestions on how we can conditionalize the build hook so it does not try to install typo3 if it's already installed? The current configuration causes `lando rebuild` to fail.

Thanks!

chadwcarlson:
@pirog 

- env vars: there aren't any specific for local that need to be set
- build hook: testing it out, it looks like the check for `platform.installed` needs to be in the `install:setup` line. After including this `lando rebuild` works fine for me.
```

        # Start the installation with no interaction or extension setup, using `SetupConfiguration.yaml`.
        if [ ! -f var/platformsh.installed ]; then
            php vendor/bin/typo3cms install:setup --install-steps-config=src/SetupConfiguration.yaml --no-interaction --skip-extension-setup
        fi;

```
@Crell

chadwcarlson:
@pirog updated in https://github.com/platformsh-templates/typo3/pull/14 & https://github.com/platformsh/template-builder/pull/365

PR #350, opened 2020-10-14
Add good Lando defaults for the magento2ce template

pirog:
@Crell and @pirog to review:

* Any MAGENTO envvars we need to set that are best for the localdevs?
* If we want users to be able to `git clone && lando start` this template then we need to update the build hook so that the existence of `.config/[env.php|config.php]` is either not assumed or is a better assumption.

This should be considered a draft for now.


sdaugherty:
Looking at the magento devdocs, there's not much that has to be touched in env.php, beyond updating the DB credentials, and allowing the dev to specify a preexisting encryption key if they are importing a DB from elsewhere.

There are some settings that might be useful there, like turning on developer mode, but IMHO we shouldn't turn those on by default (or disable caches by default) because those will result in dramatically different behavior from production, and may lead to them wasting time debugging those differences when they eventually deploy to staging.

You'll also likely want  to update base urls in your deploy hook via the magento CLI to urls that are suitable for local development

pirog:
@sdaugherty makes sense. Luckily Lando is "smart enough" to read the DB credentials, urls and keys, etc so we are good out of the box on those fronts.

The bigger concern in my mind is what settings, if any, should we set when a magento Platform.sh project is running locally on Lando. For the other templates we've configured thus far we have set things like developer mode on so your recommendation would be a break from that norm, which i think is ok since i trust your knowledge of magento much more than mine or @Crell's :) 

So let's just keep the default config and see where that gets us!

PR #354, opened 2020-10-20
add nuxtjs template

RudyWeber:
https://github.com/platformsh-templates/nuxtjs/pull/1


templates/nuxtjs/files/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Features
+
+* Node.js 14
+* Automatic TLS certificates
+* yarn-based build
+
+## Customizations
+
+The following files and additions make the framework work on Platform.sh, modified from the `npx` command [`create-nuxt-app`](https://github.com/nuxt/create-nuxt-app). If using this project as a reference for your own existing project, replicate the changes below to your project.
```
Crell, line 21: I believe Chad just changed the Gatsby or Strapi build to run that command in the platformify process, rather than having to check it into the template-builder.  Could that be done here as well?

templates/nuxtjs/files/static/sw.js
```diff
@@ -0,0 +1,17 @@
+// THIS FILE SHOULD NOT BE VERSION CONTROLLED
```
Crell, line 1: I guess that means we need to exclude it as well?

PR #357, opened 2020-10-21
README: add info for Linux install

spk:
Hi, this was needed in order to complete the install of deps because `macfsevents` is currently on the lockfile. I did not found a clean way to do that but regenerating the lock file on Linux do the trick wdyt?

cf https://github.com/pypa/pipenv/issues/3902

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ In this README we first focus on the workflow to update an existing template, th
 1. First install the dependencies:
 ```
 pipenv install
+# Note if you are on Linux please run this before to update the lock file (doit have different dependencies on Mac and Linux)
+pipenv lock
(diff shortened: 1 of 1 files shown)


Crell:
That's fascinating, because I'm on Linux and do not have to do that.

spk:
> That's fascinating, because I'm on Linux and do not have to do that.

Hum strange... It look like the packages `macfsevents` and `pyinotify` have switched hover the commits multiple times on the Pipfile.lock... Maybe its because of my pipenv version ? I'm running pipenv, version 2020.8.13

spk:
This as been fixed by https://github.com/platformsh/template-builder/pull/360

PR #360, opened 2020-10-26
Poetry instead of pipenv.

chadwcarlson:
Test with some updates: https://github.com/platformsh-templates/gatsby/pull/16


README.md
```diff
@@ -14,17 +14,24 @@ In this README we first focus on the workflow to update an existing template, th
...
 * [DoIt](http://pydoit.org/install.html)
 * [Composer](https://getcomposer.org/) (for PHP projects)
 * Git
 * Public ssh key in github account
 
 ### Setup
 
+0. Install Poetry locally
+
+Follow the installation instructions in [Poetry's documentation](https://python-poetry.org/docs/). You can verify Poetry locally with the command `poetry --version`.
```
Crell, line 27: Their instructions seem to be a bit old.  Following those instructions, I get a Poetry install that defaults to Python 2, gives me an ugly error, and then switches to Python 3.  Can we figure out how to have it install to Python 3 directly, and then include those instructions here?

README.md
```diff
@@ -67,7 +74,7 @@ Additionally, each project may have a Python class defined in the `project` dire
 
 ### Build tasks
 
-Each project has a series of build tasks, suffixed with the project name.
+Each project has a series of build tasks, suffixed with the project name. Each task can be run within a virtual environment using Poetry by running `poetry run doit` before each of the commands below. For example, `poetry run doit full:spiffy`.
```
Crell, line 77: We should also mention the `poetry shell` option here, too.

Crell:
Aside from the Python version issue, this seems to be working nicely.

PR #373, opened 2020-11-17
Rails: allow bundler to cache deps on PLATFORM_CACHE_DIR

spk:
Fix https://github.com/platformsh-templates/rails/issues/17
For this to work we will need https://github.com/platformsh/template-builder/pull/372

* Updated bundler
* Use cache dir (the build goes for the template from ~5m to ~3m)

### DIFF
--- a/templates/rails/files/.platform.app.yaml
+++ b/templates/rails/files/.platform.app.yaml
@@ -42,8 +42,21 @@ mounts:
 hooks:
     # The build hook turns what is in source control into the deployable application.
     build: |
-        gem install bundler:2.0.2
-        bundle install --without development test
+        gem install --no-document bundler:2.1.4
+        bundle config set without 'development test'
+        bundle config deployment true
+        bundle config cache_all true
+        bundle config clean true
+        echo "Installing gems"
+        # We copy the bundle directory to the Platform.sh cache directory for
+        # safe keeping, then restore from there on the next build. That allows
+        # bundler to skip downloading code it doesn't need to.
+        [ -d "$PLATFORM_CACHE_DIR/bundle" ] && \
+          rsync -az --delete "$PLATFORM_CACHE_DIR/bundle/" vendor/bundle/
+        mkdir -p "$PLATFORM_CACHE_DIR/bundle"
+        bundle install
+        # synchronize updated cache for next build
+        rsync -az --delete vendor/bundle/ "$PLATFORM_CACHE_DIR/bundle/"
         RAILS_ENV=production bundle exec rake assets:precompile
     # The deploy hook runs after your application has been deployed and started.
     deploy: |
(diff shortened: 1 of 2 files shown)


templates/rails/files/.platform.app.yaml
```diff
@@ -42,8 +42,17 @@ mounts:
...
+        bundle config set without 'development test'
+        bundle config deployment true
+        bundle config cache_all true
+        bundle config clean true
+        echo "Installing gems"
+        [ -d "$PLATFORM_CACHE_DIR/bundle-$PLATFORM_BRANCH" ] && \
+          rsync -az --delete "$PLATFORM_CACHE_DIR/bundle-$PLATFORM_BRANCH/" vendor/bundle/
+        mkdir -p "$PLATFORM_CACHE_DIR/bundle-$PLATFORM_BRANCH"
+        bundle install
+        rsync -az --delete vendor/bundle/ "$PLATFORM_CACHE_DIR/bundle-$PLATFORM_BRANCH/"
```
Crell, line 55: This block should get a comment, since it's non-obvious and these templates serve as documentation for customers with existing repos.

spk: Ok i've amended the commit feel free to change it if its not good

templates/rails/files/.platform.app.yaml
```diff
@@ -42,8 +42,19 @@ mounts:
...
     build: |
-        gem install bundler:2.0.2
-        bundle install --without development test
+        gem install --no-document bundler:2.1.4
+        bundle config set without 'development test'
+        bundle config deployment true
+        bundle config cache_all true
+        bundle config clean true
+        echo "Installing gems"
+        # when cache directory exists apply it to app so bundle use it
```
Crell, line 51:
```suggestion
        # Rails Gems don't have a reusable cache, but can benefit from having the
        # previous build available to save time. We therefore copy the downloaded gems
        # to the Platform.sh cache directory for safe keeping, then restore from there on
        # the next build.  That allows bundler to skip downloading code it doesn't need to.
```

templates/rails/files/.platform.app.yaml
```diff
@@ -42,8 +42,19 @@ mounts:
...
+        [ -d "$PLATFORM_CACHE_DIR/bundle-$PLATFORM_BRANCH" ] && \
```
Crell, line 52: `$PLATFORM_BRANCH` is not available in the build hook.  It has to be a single pool for every branch.

templates/rails/files/.platform.app.yaml
spk: Rails Gems do have reusable cache. This currently does not keep only cache directory but the whole bundler directory to avoid recompiling compiled extensions, I've pushed a modified version of your suggestion feel free to improve thanks

templates/rails/files/.platform.app.yaml
spk: I see its only on deploy too bad thanks fixed

spk:
Changes will be included in https://github.com/platformsh/template-builder/pull/390

PR #374, opened 2020-11-17
Adds Meilisearch template.

chadwcarlson:
https://github.com/platformsh-templates/meilisearch/pull/1


templates/meilisearch/files/search/build.sh
```diff
@@ -0,0 +1,27 @@
...
+
+setup_venv(){
+    echo "* SETTING UP POETRY VENV"
+    # Install poetry
+    curl https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py >> get-poetry.py
+    python get-poetry.py --version $POETRY_VERSION
+    # Source the Poetry command and cleanup.
+    . $PLATFORM_APP_DIR/.poetry/env && rm get-poetry.py
+    # Add Poetry to .bash_profile, so available during SSH.
+    echo ". $PLATFORM_APP_DIR/.poetry/env" >> ~/.bash_profile
```
RudyWeber, line 21: Shouldn't this rather go in `.environment`?

RudyWeber: Should be.
That's where we put env vars for nvm when installing an alternative node version to make this alternative version available during SSH. So this should work just the same?

templates/meilisearch/.platform.template.yaml
```diff
@@ -0,0 +1,24 @@
+version: 1
+
+info:
+  id: platformsh/meilisearch
+  name: Meilisearch
+  description: |
+      <p>This template provides a production-ready demo Meilisearch search engine on Platform.sh. The Meilisearch executable itself is downloaded during the build hook, and the example `movies` index provided in the Meilisearch documentation is added to the engine using the `meilisearch-python` library and Poetry during the deploy hook. You can specifiy the version of Meilisearch by updating the `.platform.app.yaml` file.</p>
```
RudyWeber, line 7: Mh,  if the README we say "It is not production ready, but it can be made so with only a few changes."
So we should probably not say it's production-ready here?

Issue #376, opened 2020-11-18
Update Backdrop version to 1.17.3


herbdool:
https://github.com/backdrop/backdrop/releases/tag/1.17.3

PR #377, opened 2020-11-18
Update Backdrop version to 1.17.3

herbdool:
Fixes https://github.com/platformsh/template-builder/issues/376


Crell:
Thanks.  We'll rebuild the template shortly.

PR #382, opened 2020-11-30
Apollo

RudyWeber:
https://github.com/platformsh-templates/apollo-graphql-react/pull/1

For the Front app, it'd be better if we could have https://github.com/platformsh/config-reader-nodejs/pull/17 merged.
In this example, it is vendored so I can use this feature.


templates/apollo-graphql-react/files/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+.DS_Store
+.idea/*
```
Crell, line 3: Give git its newline.

```suggestion
.idea/*

```

templates/apollo-graphql-react/.platform.template.yaml
```diff
@@ -0,0 +1,31 @@
...
+          Node.js 14<br />
+          MariaDB 10.4<br />
+          Websockets<br />
+          Automatic TLS certificates<br />
+
+initialize:
+  repository: git://github.com/platformsh-templates/apollo-graphql-react.git@master
+  config: null
+  files: []
+  profile: apollo
```
Crell, line 31:
```suggestion
  profile: Apollo and React
```

Crell, line 31: Since this is a multi-app project, we should include the flags in here to insist on a Medium plan.  See some of the other multi-app templates for what that code is. (I don't recall off hand.  I think xwiki uses it?)

templates/apollo-graphql-react/files/.platform/routes.yaml
```diff
@@ -0,0 +1,24 @@
...
+    type: upstream
+    upstream: "front:http"
+
+"https://{default}/":
+    type: redirect
+    to: "https://www.{default}/"
+
+"https://api.{default}/":
+    type: upstream
+    id: graphqlEntrypoint
```
Crell, line 16: I think IDs are supposed to be lowercase, no?  We generally use kebab case for these.

```suggestion
    id: graphql-entrypoint
```

Though I admit I detest the word "entrypoint". It's such a meaningless buzzword.

templates/apollo-graphql-react/files/.platform/services.yaml
```diff
@@ -0,0 +1,8 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+
+db:
+  type: mongodb:3.6
+  disk: 512
```
Crell, line 8: Mongo is very disk hungry.  We may want to give it a full gig out of the box, so it has room to grow.  It uses like 300 MB itself just to start up. :-/

templates/apollo-graphql-react/files/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This template shows how to setup a multi-application project exposing a GraphQL API that aggregates external APIs and internal data stored in MongoDB using Apollo Server and a client application using React, consuming the API with Apollo Client.
+
+Apollo Server is an open-source, spec-compliant GraphQL server that's compatible with any GraphQL client.<br />
+Apollo Client is state management library for JavaScript that enables you to manage both local and remote data with GraphQL.
+
+## Features
+
+* Node.js 14
+* MariaDB 10.4
```
Crell, line 17: Mongo?

templates/apollo-graphql-react/.platform.template.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  notes:
+    - heading: "Features"
+      content: |
+          Node.js 14<br />
+          MariaDB 10.4<br />
```
Crell, line 23: Mongo?

templates/apollo-graphql-react/files/back/.platform.app.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+# The relationships of the application with services or other applications.
+#
+# The left-hand side is the name of the relationship as it will be exposed
+# to the application in the PLATFORM_RELATIONSHIPS variable. The right-hand
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+  database: "db:mongodb"
+
+# The configuration of app when it is exposed to the web.
```
Crell, line 34:
```suggestion
# The configuration of the app when it is exposed to the web.
```

templates/apollo-graphql-react/files/back/src/resolvers/pubsub.js
```diff
@@ -0,0 +1,3 @@
+import { PubSub} from "apollo-server";
+
+export const pubsub = new PubSub();
```
Crell, line 3:
```suggestion
export const pubsub = new PubSub();

```

templates/apollo-graphql-react/files/back/src/resolvers/query.js
```diff
@@ -0,0 +1,55 @@
+import fetch from "node-fetch";
+
+const {
+  YOUTUBE_API_KEY = "<EXPORT YOUTUBE_API_KEY WITH YOUR KEY>",
+} = process.env;
```
Crell, line 5: If I read this correctly, it's "set the YOUTUBE_API_KEY const to the matching value out of process.env, and if that's not set then use this default message string instead."  Right?  (Just making sure my JS knowledge is up to date here.)

templates/apollo-graphql-react/files/back/src/resolvers/query.js
```diff
@@ -0,0 +1,55 @@
...
+
+export const Query = {
+  deployFridayEpisodes(
+    _parent,
+    _args,
+    { dataSources: { additionalEpisodeInfo } }
+  ) {
+    return fetch(
+      `https://youtube.googleapis.com/youtube/v3/playlistItems?maxResults=100&part=snippet&playlistId=PLn5EpEMtxTCmLsbLgaN3djvEkRdp-YmlE&key=${YOUTUBE_API_KEY}`
```
Crell, line 14: Shouldn't this URL, or at least the playlist ID, be a variable or config value somewhere?  Hard coding it seems wrong.

templates/apollo-graphql-react/files/.platform/routes.yaml
RudyWeber: The documentation doesn't specify a specific ID format, although the provided example is in kebab case, indeed. IDs being strings, it shouldn't change anything no matter what convention is used?
In JavaScript, camelCase is preferred. This mainly comes from the fact that identifiers can't contain dashes. Regarding object keys with a special characters, they can be accessed using the bracket notation: `obj["key-with-special-chars"]`, which is less convenient than the dot notation: `obj.keyWithoutSpecialChars`.
So to keep things consistent, I like having everything the same rather than wondering what typographical convention a value (that I have control over) uses.

If there are strong feelings about having ID values in camelCase, I'll change to kebab-case and adapt the code accordingly.

templates/apollo-graphql-react/files/back/src/resolvers/query.js
RudyWeber: Yup!

templates/apollo-graphql-react/files/back/src/resolvers/query.js
RudyWeber: Yeah, that'd be better. Since it's just a demo app and to keep things straightforward, there's a bunch of things I haven't done that I would normally do (e.g. proper error handling, a few abstraction layers, ...).
I also started by having a dynamic playlist ID, sent as a query variable by the front-end so any playlist could be fetched but I gave up on this idea, as it comes with other things to handle that would — I think — just add noise.

I guess it's a matter of where we want to set the good-practices cursor for targeted demos?
I will put the playlist ID in a const alongside the API key (but keep the value hard-coded). Going more into The Right Way would add more files and indirections that I think are not worth it for the sake of this example.

templates/apollo-graphql-react/.platform.template.yaml
RudyWeber: Done in a1b5856aef610548c6c6230d36eaba9757296ad7

templates/apollo-graphql-react/files/.platform/services.yaml
RudyWeber: Done in 712ae94ecda31b7fac66dabce16241644a7df1cc

templates/apollo-graphql-react/files/README.md
RudyWeber: 🤦

templates/apollo-graphql-react/files/back/src/resolvers/query.js
RudyWeber: Done in 2ee8ce95c5b449567c76267bef0ca14e706460dd
Let me know if you're not okay with this :)

PR #386, opened 2020-12-05
mount media files, sharing import files

mucan54:
Current Akeneo template doesn't mount media files, so you can not view uploaded images. Additionally if you try to import something with Akeneo importers it will not work because your uploaded files are located in app, but importer daemon works in app-queue, I added another configuration to share job files between app & app-queue with network storage.


Crell:
Thanks!

PR #389, opened 2020-12-10
Adds woocommerce and vanilla, updates bedrock and composer.

chadwcarlson:
- Bedrock copypasta: https://github.com/platformsh-templates/wordpress-bedrock/pull/2
- Composer updates: https://github.com/platformsh-templates/wordpress-composer/pull/3

- WooCommerce: https://github.com/platformsh-templates/wordpress-woocommerce/pull/1
- Vanilla: https://github.com/platformsh-templates/wordpress-vanilla/pull/1


templates/wordpress-vanilla/files/.platform.app.yaml
```diff
@@ -0,0 +1,78 @@
...
+        set -e
+
+        # Copy wp-config.php in wordpress install dir
+        rsync wp-config.php wordpress/wp-config.php
+
+        # Copy manually-provided plugins into the plugins directory.
+        # rsync -a $PLATFORM_APP_DIR/plugins/* $PLATFORM_APP_DIR/wordpress/wp-content/plugins/
+
+        # Copy manually-provided themes into the themes directory.
+        # rsync -a themes/* wordpress/wp-content/themes/
```
Crell, line 27: Why doesn't this one have $PLATFORM_APP_DIR when the plugins line does?

templates/wordpress-woocommerce/files/.environment
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+export AUTH_KEY=$PLATFORM_PROJECT_ENTROPY
+export SECURE_AUTH_KEY=$PLATFORM_PROJECT_ENTROPY
+export LOGGED_IN_KEY=$PLATFORM_PROJECT_ENTROPY
+export NONCE_KEY=$PLATFORM_PROJECT_ENTROPY
+export AUTH_SALT=$PLATFORM_PROJECT_ENTROPY
+export SECURE_AUTH_SALT=$PLATFORM_PROJECT_ENTROPY
+export LOGGED_IN_SALT=$PLATFORM_PROJECT_ENTROPY
+export NONCE_SALT=$PLATFORM_PROJECT_ENTROPY
```
Crell, line 23: Is the extra intermediary step of going through extra env vars worth it?  The entropy variable doesn't need JSON decoding like relationships do.  Why not just read that with `getenv()` in the config file and read it directly?  (Applies to all WP templates here.)

templates/wordpress-vanilla/files/wp-config.php
```diff
@@ -0,0 +1,98 @@
...
+				'AUTH_KEY',
+				'SECURE_AUTH_KEY',
+				'LOGGED_IN_KEY',
+				'NONCE_KEY',
+				'AUTH_SALT',
+				'SECURE_AUTH_SALT',
+				'LOGGED_IN_SALT',
+				'NONCE_SALT',
+			];
+			$entropy = getenv('PLATFORM_PROJECT_ENTROPY');
```
Crell, line 51: Like... exactly what this code is doing. This seems better than trying to set the env vars based on it as well.  Or am I missing something?

templates/wordpress-vanilla/files/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Your config, not a part of WP upstream
+wordpress/wp-config.php
+
+# Custom plugins
+# wordpress/wp-content/plugins/classic-editor
+# wordpress/wp-content/plugins/woocommerce
```
Crell, line 6: Why are these commented out ones here?

Crell: Also, in vanilla, shouldn't they be committed as normal rather than any fancy download and copy system?

templates/wordpress-woocommerce/files/.lando.upstream.yaml
```diff
@@ -0,0 +1,17 @@
...
+# These both allow you to test this template without needing a site on Platform.sh
+# However you will want to replace them in your .lando.yml
+name: platformsh-woocommerce
+recipe: platformsh
+
+# These are tools that are commonly used during development for this template.
+tooling:
+  wp:
+    service: app
+
```
Crell, line 17: Keep Git happy about whitespace.

templates/wordpress-woocommerce/files/.platform.app.yaml
```diff
@@ -0,0 +1,69 @@
...
+mounts:
+    "web/app/uploads":
+        source: local
+        source_path: "app-uploads"
+    "web/wp/wp-content/cache":
+        source: local
+        source_path: "cache"
+    "web/wp/wp-content/uploads":
+        source: local
+        source_path: "uploads"
```
Crell, line 69: Keep Git happy about whitespace, the other direction. :smile:

templates/wordpress-woocommerce/files/.platform/services.yaml
```diff
@@ -0,0 +1,9 @@
+# The services of the project.
+#
+# Each service listed will be deployed
+# to power your Platform.sh project.
+
+db:
+    type: mariadb:10.4
+    disk: 2048
+
```
Crell, line 9: Git is really picky about its whitespace.

templates/wordpress-woocommerce/files/.environment
chadwcarlson: WooCommerce is built on the Bedrock template. You can do both, but Bedrock's big sell is environment variables straight away.

templates/wordpress-vanilla/files/.gitignore
chadwcarlson: So my thinking here is
- commit WP to `wordpress`
- commit themes and plugins to subdirs outside of `wordpress`, same way we're recommending using non-Composer plugins in the [deploy guide](https://docs.platform.sh/guides/wordpress/deploy/next-steps.html#adding-plugins-and-themes-without-composer)
- at build everythings moves into `wordpress` to run like you'd expect, but then you can update in each separate subdir or update as submodule however you'd like there.
- the gitignore lines are for when running Lando and the build hook is run, left as only an example but I'll remove.

templates/wordpress-woocommerce/files/.environment
Crell: Ahhh... It already has the env var glue built in?  Good for them. :smile: OK, stet.

templates/wordpress-vanilla/files/.gitignore
Crell: If you're doing vanilla WP, then you have no local build process.  As setup here, it won't work in local development because all of the modules are in the wrong place.

Vanilla WP should be, well, vanilla.

chadwcarlson: I guess in this case I'm pushing local Lando, so users can keep all of it up to date the easiest way - whether thats downloading new versions straight away or updating submodules. Makes sense to me since we're recommending it that way already with the deploy guide above

templates/wordpress-vanilla/files/.platform.app.yaml
```diff
@@ -0,0 +1,78 @@
...
+        set -e
+
+        # Copy wp-config.php in wordpress install dir
+        rsync wp-config.php wordpress/wp-config.php
+
+        # Copy manually-provided plugins into the plugins directory.
+        # rsync -a $PLATFORM_APP_DIR/plugins/* $PLATFORM_APP_DIR/wordpress/wp-content/plugins/
+
+        # Copy manually-provided themes into the themes directory.
+        # rsync -a $PLATFORM_APP_DIR/themes/* $PLATFORM_APP_DIR/wordpress/wp-content/themes/
```
Crell, line 27: These can go away.

PR #398, opened 2021-01-22
Add email support to Directus.

chadwcarlson:
https://github.com/platformsh-templates/directus/pull/3

Also removes the stray `https:` auth directory from template-builder, and adds Redis and memcached support to the template.

### DIFF
--- a/templates/directus/files/.platform.app.yaml
+++ b/templates/directus/files/.platform.app.yaml
@@ -16,12 +16,13 @@ type: nodejs:12
 # side is in the form `<service name>:<endpoint name>`.
 relationships:
     database: "dbpostgres:postgresql"
+    rediscache: "cacheredis:redis"
+    memcachedcache: "cachemc:memcached"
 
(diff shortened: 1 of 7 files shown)


templates/directus/files/.platform.app.yaml
```diff
@@ -20,7 +20,15 @@ relationships:
 # The hooks that will be triggered when the package is deployed.
 hooks:
     build: |
+        # Use module's example.env file as .env.
         cp node_modules/directus/example.env .env
```
Crell, line 24: Ha!  I literally just wrote a blog post saying to not use `.env` on Platform.sh. :smile: 

Why is it doing this?

chadwcarlson: Totally get you, but seems like this whole thing breaks if a `.env` is not committed, actually. Investigating...

chadwcarlson: Ope, nevermind. Figured it out. Update on the way

PR #409, opened 2021-02-25
Add cache mount to Nextjs

chadwcarlson:
https://github.com/platformsh-templates/nextjs/pull/5


templates/nextjs/files/.platform.app.yaml
```diff
@@ -20,3 +20,13 @@ hooks:
...
+
+# The size of the persistent disk of the application (in MB).
+disk: 512
+
+# The mounts that will be performed when the package is deployed.
+mounts:
+    # Next.js will try to cache files, so it must be writeable.
+    '/.next/cache':
+        source: local
+        source_path: 'next-cache'
```
RudyWeber, line 32: For SSR / ISR, we will also need `.next/server`.

```suggestion
    '/.next/cache':
        source: local
        source_path: 'next-cache'
    '.next/server':
        source: local
        source_path: 'next-server'
```

RudyWeber (changes requested):
We should also make sure not to keep the content of `.next/cache` and `.next/server`.
Next generates cache and build files that are re-used at runtime for images / SSR / ISR / ...

RudyWeber (approved):

Issue #468, opened 2021-12-07
composer templates `--ignore-platform-req=php` allows installation of incompatible packages


gilzow:
For composer-based templates, `--ignore-platform-req=php` is added a default composer parameter to allow for anyone to run the template update process regardless of what they have installed on their local machine. Unfortunately, by ignoring the php version, we allow composer to install versions of dependencies that may be incompatible with the version of php defined in .platform.app.yaml. Instead, we need to utilize `config:platform:php` to instruct composer to use that version for installing/updating.

PR #469, opened 2021-12-07
Wordpress/composer based fix

gilzow:
Addresses #468 partially. This PR fixes the issue for composer-based WordPress templates so we can fix a collection of issues in WordPress templates. We'll then expand this fix out to other composer-based templates using a similar structure. 

**Adds**
*  `.type` and `.typeVersion` attributes to `BaseProject` class by reading .platform.app.yaml and extracting the `type` key from the yaml file
* `any_constructor` method to `BaseProject` class to allow parsing yaml files that contain `!tags` (e.g. `!include`) that would normally prevent pyyaml from reading the file
* override `composer_defaults` method in class `WordPressComposerBase` to remove `--ignore-platform-req=php` composer param from composer defaults
* override `platformify` method in class `WordPressComposerBase` to allow us to inject `composer config platform.php {self.typeVersion}` before the rest of the composer commands, and `composer config --unset platform` after all the composer commands have run
* pyyaml ^5.4.1 as dependency of project, regens poetry.lock to include new dependency.  

**Updates**
* renames method `platformify()` to `_platformify()` in all child classes of class `WordPressComposerBase`
* autopep8 1.5.4 --> 1.6.0
* pylint 2.6.0 --> 2.12.2

**Fixes**
* Structure of child classes of WordPressComposerBase and how they provide the list of dependencies to unpin


gilzow:
Hold on this one. I almost have a complete fix for #468 ready

gilzow:
**Updates**  
- All template child classes to include decorated platformify method. **See notes below.**

**Adds**   
- New intermediary `Composer` class between `RemoteProject` class and the template classes that use composer for dependency management (see notes below):
    - Extracts php version from `.platform.app.yaml` and sets it as `self.composer_version`
    -  adds @ComposerProject.composer_platformify decorator to set composer config:platform.php` to match version from `.platform.app.yaml` before running `composer update` and removes it once we're finished
    - Extraction of `dependencies:php:composer/composer` to set template.composer_version. if not available, sets to 1 as default.
    - Logic for determining where to inject the extra commands in the decorator using the name of the function being decorated (ie update needs our command to be injected after `git checkout` has occurred vs `platformify` where it can be the very first command)

**Removes**   
- All the composer-related functionality in `__init__.py` and adds it to `composer.py`.  **See notes below**

**Notes**  
The Akeneo template and related python class require further investigation before being converted to use the new composer class. Therefore, some of the composer-related functionality in `__ini__.py` that it relies on has been left in place until Akeneo can be converted.

PR #522, opened 2022-02-17
Alters Step 4 to use the platformsh/prep-for-autopr@main github action

gilzow:
Bulk of step 4 was moved into a reusable github action. updates the workflow to use this action.


gilzow:
In testing the changes, I discovered an issue in Step 7 when we check to see if origin already has an update branch, if it is related to the default branch, and if so, if it has advanced beyond when it was branched from master. For some reason, our local clone of the repo was not up-to-date with origin and didn't have a record of the branches. Added a `git fetch origin` before attempting to check its history.

chadwcarlson (approved):

PR #532, opened 2022-02-22
Testing non-included templates

chadwcarlson:


chadwcarlson:
akeneo & drupal8 moved a bit further, but not yet working perfectly.

chadwcarlson:
- sculpin needs more investigation
- directus breaks when juggling mounts. Is this really needed anymore?

chadwcarlson:
Magento: needs to be updated to match pure composer version https://github.com/jww-sh/magento2ce/

PR #570, opened 2022-03-16
update nextjs files

hacktivist123:
This PR updates the NextJS template to prevent it from ending up with orphan” processes.


hacktivist123:
https://github.com/platformsh-templates/nextjs/pull/45

PR #592, opened 2022-04-20
update flask template

hacktivist123:
- Update the ReadME with more information 
- Update Flask to v2
- Specify a newer Python version
- Add a basic Blackfire config file
- Switch pipenv to poetry


hacktivist123:
https://github.com/platformsh-templates/flask/pull/15

PR #610, opened 2022-05-25
Adds NextJS-WordPress template doc

gilzow:
Per conversation, adds NextJS+WordPress template doc to be able to list in console.


templates/nextjs-wordpress/.platform.template.yaml
```diff
@@ -0,0 +1,60 @@
...
+    # A list of tags associated with the template.  These should be highly generic terms like "CMS", "Framework", and
+    # the language in which the application is written.
+    tags:
+        - PHP
+        - CMS
+        - Node.js
+        - Decoupled
+        - Multi-app
+    # An image URI (either base64-encoded or a URL) representing the template.  Base64-encoded SVG strongly preferred.
+    image: data:image/svg+xml;...
```
chadwcarlson, line 40: Looks good, only thing missing here is the image for when it appears in console's catalog.

https://github.com/platformsh/template-builder/blob/ff1bb243bd58817da51ab05b0f61c092efdf49e5/templates/wordpress-composer/.platform.template.yaml#L13

PR #614, opened 2022-05-27
Update wordpress-bedrock .environment file

gilzow:
in the [.environment file](https://github.com/platformsh-templates/wordpress-bedrock/blob/master/.environment), it's setting the [constant WP_ENV to development if the current branch isn't master](https://github.com/platformsh-templates/wordpress-bedrock/blob/master/.environment#L10). However, most have changed their default/production away from that term to something else which means every branch has WP_ENV set to development.

In addition, now that we have PLATFORM_ENVIRONMENT_TYPE with values of production|development|staging we should be able to change the codeblock in question to just

This will close #48 @ https://github.com/platformsh-templates/wordpress-bedrock/issues/48

### DIFF
--- a/templates/wordpress-bedrock/files/README.md
+++ b/templates/wordpress-bedrock/files/README.md
@@ -28,7 +28,7 @@ The following changes have been made relative to Bedrock's project creation comm
 * The `.platform.app.yaml`, `.platform/services.yaml`, and `.platform/routes.yaml` files have been added.  These provide Platform.sh-specific configuration and are present in all projects on Platform.sh.  You may customize them as you see fit.
 * A `.environment` file has been included which sets Bedrock's configuration on Platform.sh. It defines the connection to the MariaDB database, the primary site URL, and security keys and salts.
 * A base [Landofile](https://docs.lando.dev/config/lando.html#base-file) provides configuration to use this template locally using [Lando](https://docs.lando.dev).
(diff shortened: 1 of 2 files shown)


templates/wordpress-bedrock/files/README.md
```diff
@@ -26,8 +26,9 @@ Run through the WordPress installer as normal. You will not be asked for databas
...
 
 * The `.platform.app.yaml`, `.platform/services.yaml`, and `.platform/routes.yaml` files have been added.  These provide Platform.sh-specific configuration and are present in all projects on Platform.sh.  You may customize them as you see fit.
-* A `.environment` file has been included which sets Bedrock's configuration on Platform.sh. It defines the connection to the MariaDB database, the primary site URL, and security keys and salts. 
+* A `.environment` file has been included which sets Bedrock's configuration on Platform.sh. It defines the connection to the MariaDB database, the primary site URL, and security keys and salts.
 * A base [Landofile](https://docs.lando.dev/config/lando.html#base-file) provides configuration to use this template locally using [Lando](https://docs.lando.dev).
+* In order to preserve 1-to-1 clones of development environments, `WP_ENV` has been set to `production` in [`.environment`](./.environment). If you would instead prefer to have `WP_ENV` match the type of platform.sh environment (i.e. development, staging), you can set change that line to `export WP_ENV="${PLATFORM_ENVIRONMENT_TYPE}"`.
```
chadwcarlson, line 31:
```suggestion
* In order to preserve 1-to-1 clones of development environments, `WP_ENV` has been set to `production` in [`.environment`](./.environment). If you would instead prefer to have `WP_ENV` match the type of Platform.sh environment (i.e. development, staging), you can set change that line to `export WP_ENV="${PLATFORM_ENVIRONMENT_TYPE}"`.
```

chadwcarlson (approved):

PR #626, opened 2022-06-27
update gatsby-strapi template

hacktivist123:
In this PR I updated the Gatsby Strapi template to an entirely new and improved version to match Strapi 4 and Gatsby 4 requirements. It comes with a fresh ReadME to make sure it is used correctly.


hacktivist123:
https://github.com/platformsh-templates/gatsby-strapi/pull/14

PR #654, opened 2022-08-24
Adds ?per_page=100 to the github api call 

### LINKED ISSUE #652: Github release list api limits results to 30, breaking mattermost
In the Mattermost project, we need to [retrieve the list of releases](https://github.com/platformsh/template-builder/blob/master/project/mattermost.py#L19) that match what we have set as `major_version` (which, at the time of this writing is 6.1). [According to the docs](https://docs.github.com/en/rest/releases/releases#list-releases--parameters), the default number of releases returned from this api endpoint is 30. At some point recently, enough new releases have been made to push all 6.1.X releases past 30 which causes any template builds to fail as we are unable to retrieve releases in the mattermost project.

_For now_, I suggest adding `?per_page=100` to the API call to get us over the hump. _In the near future_, we will need to expand this section to address pagination.

gilzow:
Adds ?per_page=100 to the github api call so we can retrieve enough results in order to find the v6 tags

Closes #652

### DIFF
--- a/project/mattermost.py
+++ b/project/mattermost.py
@@ -3,6 +3,7 @@ import os
 import json
 import requests
 
+
 class Mattermost(BaseProject):
 
     @property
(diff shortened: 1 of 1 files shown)


chadwcarlson:
@gilzow mind tacking this same thing onto Hugo? https://github.com/platformsh/template-builder/blob/master/project/hugo.py

PR #670, opened 2022-09-28
Update templates tags to improve the template picker

lolautruche:


templates/django4/.platform.template.yaml
```diff
@@ -29,6 +29,7 @@ info:
   tags:
   - Python
   - Framework
+  - Django
```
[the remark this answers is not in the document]
lolautruche: Yes, similar to Symfony, or other PHP frameworks. This is for the template picker to get the appropriate filters.

templates/typo3/.platform.template.yaml
```diff
@@ -9,6 +9,7 @@ info:
   tags:
   - PHP
   - CMS
+  - Typo3
```
flovntp, line 12: As for Sylius, Typo3 is using Symfony (components only) so i don't know if we need to make SF appears here... 
if yes, it should appears also for Drupal templates, right ?

lolautruche: Ah I wasn't aware of this. Yes, Symfony might appear here as well, but I'm not sure this makes sense (same for Drupal)

lolautruche: Sylius is Symfony full stack, Drupal and Typo3 are not

flovntp: true, so lets forget about all those comments...

chadwcarlson (approved):

Issue #692, opened 2022-11-25
Commands require all templates to be loaded


chadwcarlson:
Running `poetry run doit full:drupal9` still loads every template's configuration into doit for that command, even though it isn't run. That causes situations where a single template (i.e. Hugo) has some bug in how it grabs upstream versions, which will cause an unrelated template update to fail.

With a little testing, it looks like we can make the following modification to each doit task to prevent this from happening:

```python
def task_check():
    # Grab the incoming arg, which in this case will be 'check:drupal9', then split it up to filter.
    template_name=sys.argv[1].split(':')[1]
    for project in ALL_PROJECTS:
        if project.name == template_name:
            yield {
                'name': project.name,
                'actions': project.check,
            }
```

Issue #693, opened 2022-11-28
Update cron syntax


chadwcarlson:
Testing https://github.com/platformsh/template-builder/issues/692, Drupal 9 fails under the old cron syntax. 

Update templates to use the cron definition dictionary https://docs.platform.sh/create-apps/app-reference.html#cron-commands

gilzow:
Closed by #694

PR #694, opened 2022-11-28
Cron fix, single template pass.

chadwcarlson:
Fixes https://github.com/platformsh/template-builder/issues/692

Provides an update to Drupal 9 for https://github.com/platformsh/template-builder/issues/693, which will need to be applied to all templates that have crons defined. 

There's also a stray link output into the Drupal 9 README file that's removed.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -6,6 +6,7 @@ In case the actions of some tasks need to be customized, the new BaseProject sub
 '''
 
 import os
+import sys
 
 from project import BaseProject, TEMPLATEDIR
 from project.akeneo import Akeneo
@@ -63,11 +64,13 @@ def task_cleanup():
 
     Usage: doit cleanup:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
     for project in ALL_PROJECTS:
-        yield {
-            'name': project.name,
-            'actions': project.cleanup,
-        }
+        if project.name == template_name:
+            yield {
+                'name': project.name,
+                'actions': project.cleanup,
+            }
 
 
 def task_init():
@@ -76,12 +79,14 @@ def task_init():
 
     Usage: doit init:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
     for project in ALL_PROJECTS:
-        yield {
-            'name': project.name,
-            'task_dep': ['cleanup:{0}'.format(project.name)],
-            'actions': project.init,
-        }
+        if project.name == template_name:
+            yield {
+                'name': project.name,
+                'task_dep': ['cleanup:{0}'.format(project.name)],
+                'actions': project.init,
+            }
 
 
 def task_update():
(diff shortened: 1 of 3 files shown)


flovntp:
it seems fine for me 👍

dodo.py
```diff
@@ -63,11 +64,13 @@ def task_cleanup():
 
     Usage: doit cleanup:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
```
gilzow, line 67: If the project name is left out (ie meaning we want to run the command against all projects), then line 67 throws an exception since `1` is our of range of the list.

dodo.py
```diff
@@ -117,11 +126,13 @@ def task_branch():
 
     Usage: doit branch:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
```
gilzow, line 129: same

dodo.py
```diff
@@ -130,11 +141,13 @@ def task_push():
 
     Usage: doit push:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
```
gilzow, line 144: same

dodo.py
```diff
@@ -143,14 +156,16 @@ def task_rebuild():
 
     Usage: doit rebuild:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
```
gilzow, line 159: same

dodo.py
```diff
@@ -159,11 +174,13 @@ def task_full():
 
     Usage: doit full:<project>
     """
+    template_name=sys.argv[1].split(':')[1]
```
gilzow, line 177: one more

gilzow (commented):
We should probably decide if we still want to support ALL project commands, or if we want to enforce running template builder against individual projects.

chadwcarlson:
Suggestion,

Update 

```py
    ALL_PROJECTS = [project_factory(f.name) for f in os.scandir(TEMPLATEDIR)
                    if f.is_dir() and f.name not in IGNORED]
```

to 

```py
if ':' in sys.argv[1]:
    ALL_PROJECTS = [project_factory(f.name) for f in os.scandir(TEMPLATEDIR)
                    if f.is_dir() and f.name not in IGNORED and f.name == sys.argv[1].split(':')[1]]
else:
    ALL_PROJECTS = [project_factory(f.name) for f in os.scandir(TEMPLATEDIR)
                    if f.is_dir() and f.name not in IGNORED]
```

so individual and ALL projects are still supported with a smaller change.

chadwcarlson:
Updated on above, tested here https://github.com/platformsh-templates/nextjs/pull/85

dodo.py
```diff
@@ -55,22 +55,25 @@ def project_factory(name):
     except KeyError:
         return BaseProject(name)
 
-ALL_PROJECTS = [project_factory(f.name) for f in os.scandir(TEMPLATEDIR)
-                if f.is_dir() and f.name not in IGNORED]
+
+if ':' in sys.argv[1]:
```
gilzow, line 59: don't we need to `import sys` ?

gilzow (changes requested):
~~Just need to `import sys` and then this should be good to merge~~
Disregard. I was looking at your last commit and didn't see it. I see now that it was in an earlier commit. My apologies.

PR #733, opened 2022-12-13
Resyncs files and structures for the gatsby-drupal template

### LINKED ISSUE #730: Bug: Resync Gatsby-Drupal template
The [Gatsby Drupal template](https://github.com/platformsh-templates/gatsby-drupal) has Drupal installed to an [api](https://github.com/platformsh-templates/gatsby-drupal/tree/master/api) directory and Gatsby in a [client](https://github.com/platformsh-templates/gatsby-drupal/tree/master/client) directory. However, the structure for the Gatsby Drupal template [here in template-builder](https://github.com/platformsh/template-builder/tree/master/templates/gatsby-drupal/files) has Drupal in a `drupal` directory, and Gatsby in a `gatsby` directory. Attempting to update the Gatsby Drupal template from template builder results in 4 directories in the build: api, client, drupal, and gatsby. 

We need to update template builder to match the template.

gilzow:
Resyncs files and structures for the gatsby-drupal template to match what is in the [platformsh-templates/gatsby-drupal ](https://github.com/platformsh-templates/gatsby-drupal) repository. Closes #730 

Also updates cron syntax in api/.platform.app.yaml.

There are a few things that I did **not** bring over from the current platformsh-templates/gatsby-drupal repository:
* `.ddev` directory and its subfolders. This should no longer be needed now that we have the ddev-platformsh addon, correct?
* an [.htaccess file](https://github.com/platformsh-templates/gatsby-drupal/blob/master/api/config/sync/.htaccess) that is in [api/config/sync](https://github.com/platformsh-templates/gatsby-drupal/tree/master/api/config/sync). This is not needed, correct? 

Lastly, the current gatsby-drupal repo does not include the [install-redis.sh ](https://github.com/platformsh/template-builder/blob/master/templates/gatsby-drupal/files/drupal/install-redis.sh)file. I have not removed this file _yet_ from template builder. This is no longer needed, correct?


Mupsi:
I don't know if the add-on means we don't need the .ddev folder anymore, that's a good question. I'd keep it for now, just to be safe, as not everybody may have heard of the add-on yet.
I have to info about the .htaccess either, sorry! I don't see any mention about it in the official doc for the Drupal Gatsby module :woman_shrugging:

chadwcarlson:
@gilzow 

- `install-redis.sh` is no longer needed
- correct on `.ddev` 
- This would be a question for DX - without that file present, you get a `.htaccess` file missing error message that takes up a good amount of space in the logs. Thoughts?

gilzow:
> `install-redis.sh` is no longer needed ✅ 

> correct on `.ddev` ✅ 

> This would be a question for DX - without that file present, you get a .htaccess file missing error message that takes up a good amount of space in the logs. Thoughts?

If it's going to cause errors or log entries, then we should leave it. It's not _hurting_ anything by staying in place. I made the incorrect assumption that by being an .htaccess file, and not usable in our system, that it should be removed.

chadwcarlson (approved):

PR #793, opened 2023-02-13
adds source ops auto-update to .platform.app.yaml for FastAPI template

gilzow:
see title


chadwcarlson:
Included in https://github.com/platformsh/template-builder/pull/794

PR #830, opened 2023-04-06
drupal10: don't require script to be executable

pjcdawkins:
It's awkward to call scripts like this inside templates, as customers may have copied the files without copying the permissions.


gilzow:
huh. I wonder if this was a carry-over from the D9 template? you're correct, we shouldnt be assuming the permissions are correct. 

Right now template-builder --> D10 template updates are paused, so I'll make the change there as well.

PR #835, opened 2023-04-12
Fix git repository route protocol

ptltbt:
The current version of the template points to the Typo3-v11 repo using the `git` protocol. This causes an error when Platform.sh tries to clone the repository. Changing it to `https` resolves this.


gilzow (commented):
correct, this should have been https

gilzow (approved):

PR #850, opened 2023-05-08
Skip nuxtjs updates until synced.

chadwcarlson:
To be merged alongside https://github.com/platformsh-templates/nuxtjs/pull/72

- skips workflow until synced.
- Updates template.yaml for catalog


gilzow (approved):
Makes sense

Issue #825, opened 2023-03-16
Template updating improvements


gilzow:
Next steps to be taken in improvements to the template updating process. cc: @chadwcarlson 

- [ ] Migrate current set of priority templates from using auto-updating workings to use reusable workflows
    - [ ] Add reusable workflow based on [@chadwcarlson 's POC](https://github.com/chadwcarlson/fastapi/blob/master/.github/workflows/tb-sync.yaml) to monitor changes to PSH config files and push those changes as a PR to this repository for the affected template
    - [ ] Replace the [current workflows](https://github.com/platformsh/template-builder/tree/master/common/sourceop-auto-update/.github/workflows) in this repo with the new reusable workflow versions 
- [ ] For the Priority Templates, continue separating what is tracked in template-builder vs what is tracked in a template's repository
    - This refers to having template builder track only PSH configuration files for a template while all other parts are tracked in the template's repository
    - [ ] Look into the Remote Project issue that might be surfaced when this migration is complete.
- [ ] Add in the type of template (e.g. Starters, Demos, Experimental) into each template's `.platform.template.yaml` file
    - [ ] Discover which property in the above file should be used for this information
- [ ] Look into adding a new doit command that creates a new template adhering to this new methodology

chadwcarlson:
## Template reliability roadmap

1. Set up templates issue/PR GitHub Project
   - [ ] Template-builder board
   - [ ] "All templates" board (to start)

2. Finalize the auto-update (daily) and template-builder update (monthly?) workflow for **Starters**
   - [ ] Migrate current set of priority templates from using auto-updating workflows to use reusable workflows
   - [ ] Add reusable workflow based on [@chadwcarlson 's POC](https://github.com/chadwcarlson/fastapi/blob/master/.github/workflows/tb-sync.yaml) to monitor changes to PSH config files and push those changes as a PR to this repository for the affected template
   - [ ] Replace the [current workflows](https://github.com/platformsh/template-builder/tree/master/common/sourceop-auto-update/.github/workflows) in this repo with the new reusable workflow versions
   - [ ] Add in the type of template (e.g. Starters, Demos, Experimental) into all template's .platform.template.yaml file ([See below](#template-classes)

3. Eliminate community contribution friction on templates

   - [ ] For the Priority Templates (**Starters**), continue separating what is tracked in template-builder vs what is tracked in a template's repository. This refers to having template builder track only PSH configuration files for a template while all other parts are tracked in the template's repository
[...]

PR #889, opened 2023-07-27
Enhance/846b tb cleanup phase1dot5

### LINKED ISSUE #846: [Enhancement] Phase 1 of template clean-up
See #825 
Phase 1 involves removing all non-psh config files for a single Starter template (WordPress-Composer) and running through the whole template-builder process to ensure we have everything accounted for. Given the diversity of our templates, we want to make several progressive changes to the process to ensure we don't miss anything

gilzow:
Closes #846  and partially completes #825 
This is a big one. For 27 templates designated as starters
* it removes all non-platform.sh configuration files from their areas in template builder. For those removed files, management and tracking of those files are moved to their individual repositories. 
* platform.sh configuration files have been resynced with the versions in their individual templates as they have drifted over time
* platform.sh configuration files have had their internal properties re-ordered so that the list of properties matches the order in the documentation 
* platform.sh configuration files have had internal documentation added to each property that includes a brief explanation and a link to the relevant docs for more/complete information 
* for the above templates that have a remote source (e.g. Drupal10), new functionality has been added as a mixin to address the renaming of the upstream README (to README_UPSTREAM) so it doesn't clobber our README that we need in the templates' repository
* Additional clean-up was made as discovered (correcting cron syntax, updating template builder commands for a template where things have changed, updating runtimes, etc)
* Pushes out new github workflows to all the above templates that enables auto-updating via source operations


project/ReadmeMixin.py
```diff
@@ -0,0 +1,28 @@
+class ReadmeMixin:
+    readMeFileName = "README"
+    readMeFileExt = "md"
+    readMeUpstreamFile = "_UPSTREAM"
+    readMePSHFile = "_PSH"
+
+    def rename_psh_readme(self):
```
chadwcarlson, line 7: Mixin!

PR #900, opened 2024-02-06
Sync: matching platformsh-templates/magento2ce

platformsh-devrel:
Syncing updates for template platformsh/template-builder made in the latest [pull request](https://github.com/platformsh-templates/magento2ce/commit/a70f58668f4b98cdec3e4ace3f9ce28b53412eb9).


gilzow:
This is significantly different, and I'm not sure why. we need to research why these two have drifted so far apart.

gilzow (approved):

PR #914, opened 2025-04-22
fix: commands expect `start` definition

spk:
cf https://docs.platform.sh/create-apps/upgrading.html#changes-in-version-202202


flovntp (approved):

flovntp:
cc @gilzow as you're my template builder's expert ;)

gilzow:
for those that aren't archived, we need to update the source repositories:
* [Akeneo](https://github.com/platformsh-templates/akeneo)
* [Laravel](https://github.com/platformsh-templates/laravel)
* [Nextcloud](https://github.com/platformsh-templates/nextcloud)

All of the rest are publicly archived.

